PR #45, opened 2020-05-13
To fix the issue where loopback address wasn't populated as expected via ios_l3_interfaces module

justjais:
##### SUMMARY

To fix the issue where loopback address wasn't populated as expected via ios_l3_interfaces module, fixes issue raised via #43

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ios_l3_interfaces

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/network/ios/config/l3_interfaces/l3_interfaces.py
+++ b/plugins/module_utils/network/ios/config/l3_interfaces/l3_interfaces.py
@@ -18,14 +18,13 @@ from ansible_collections.ansible.netcommon.plugins.module_utils.network.common.c
 )
 from ansible_collections.ansible.netcommon.plugins.module_utils.network.common.utils import (
     to_list,
-    remove_empties
 )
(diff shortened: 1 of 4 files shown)


justjais:
@GomathiselviS I have updated the merge TC to check for the scenario

justjais:
recheck

plugins/module_utils/network/ios/config/l3_interfaces/l3_interfaces.py
```diff
@@ -141,7 +141,7 @@ def _state_replaced(self, want, have, module):
 
         for interface in want:
             for each in have:
-                if each["name"] == interface["name"]:
+                if each["name"].lower() == interface["name"].lower():
```
NilashishC, line 144: Wouldn't this get solved by passing the want through normalize_interface() before calling set_state()? Seems more robust than having to deal with this everywhere separately.

justjais: yea, make sense

PR #56, opened 2020-05-22
Fix ios_vlans replace operation with non-existing vlan config

justjais:
##### SUMMARY

Fix ios_vlans replace operation with non-existing VLAN config as with non-existing VLANs replace operation was skipping and showing changed as False which was ideally incorrect. Vlans config should be fired even if the VLANs don’t pre-exist with replace operation as well. Both merge and override operation is behaving as expected.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ios_vlans
##### ADDITIONAL INFORMATION

```paste below

```


justjais:
recheck

PR #90, opened 2020-06-23
Add changelog files

ganeshrn:
##### SUMMARY

*  Add changelog file for 1.0.0 release genereated
   using https://github.com/ansible-community/antsibull-changelog

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

changelogs

##### ADDITIONAL INFORMATION

```paste below

```


changelogs/changelog.yaml
```diff
@@ -0,0 +1,108 @@
...
+    - description: Interfaces resource module
+      name: ios_interfaces
+      namespace: ''
+    - description: (deprecated, removed after 2022-06-01) Manage Layer-2 interface
+        on Cisco IOS devices.
+      name: ios_l2_interface
+      namespace: ''
+    - description: L2 interfaces resource module
+      name: ios_l2_interfaces
+      namespace: ''
```
justjais, line 39: is there a specific need to keep the namespace, if we're not populating the same?

ganeshrn: Yes, it is for the nested folder in the modules directory for eg the nxos collection has modules under the storage directory. If the nested folder is not present the changelog tool expects namespace value to `''`

PR #107, opened 2020-07-14
Make config modules work properly when module alias is used in task

NilashishC:
Signed-off-by: NilashishC <nilashishchakraborty8@gmail.com>

##### SUMMARY
- Module name will not always be prefixed with {{ network_os }}_.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
action/ios.py
ios_config.py


NilashishC:
recheck

PR #114, opened 2020-07-22
To remove IOS ignores from sanity test

justjais:
##### SUMMARY

To remove IOS ignores from sanity test

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ios

##### ADDITIONAL INFORMATION

```paste below

```


plugins/modules/ios_l2_interface.py
```diff
@@ -398,7 +448,7 @@ def main():
     """ main entry point for module execution
     """
     element_spec = dict(
-        name=dict(type="str", aliases=["interface"]),
+        name=dict(type="str", aliases=["interface"], required=True),
```
rohitthakur2590, line 451: top-level name cannot be required for modules having the aggregate key

plugins/modules/ios_vlan.py
```diff
@@ -324,10 +370,10 @@ def main():
     """ main entry point for module execution
     """
     element_spec = dict(
-        vlan_id=dict(type="int"),
+        vlan_id=dict(type="int", required=True),
```
rohitthakur2590, line 373: top-level name cannot be required for modules having the aggregate key

Issue #248, opened 2021-02-10
l2_interfaces.py _check_for_correct_vlan_range is comparing string instead of int


pierremahot:
##### SUMMARY
l2_interfaces.py _check_for_correct_vlan_range is comparing string instead of int

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
l2_interfaces.py _check_for_correct_vlan_range

##### ANSIBLE VERSION

```
ansible 2.10.5
  config file = /roo/ansible-skeleton/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version 3.8.5  (default, Jul 28 2020, 12:59:40) [GCC 9.3.0]
```

##### OS / ENVIRONMENT
for ansible :
Ubuntu 20.04
for the switch :
cisco C2960 ios 15.02

##### STEPS TO REPRODUCE
```yaml
- name: add vlan to trunk allowed
  cisco.ios.ios.ios_l2_interfaces:
    config:
      - name: "GigabitEthernet1/0/1"
        trunk:
          allowed_vlans: ["3-12"]
  become: yes
```

##### EXPECTED RESULTS
no error

##### ACTUAL RESULTS
```
Command rejected: Bad VLAN list - end of range not larger than the start of range!
```
I have modify the msg to get the two values and message is : 
```
Command rejected: Bad VLAN list - end of range not larger than the start of range! 3 >= 12
```

justjais:
@pierremahot PR is merged, thus closing the issue.

PR #249, opened 2021-02-10
bug #248 resolved

pierremahot:
FIX issue https://github.com/ansible-collections/cisco.ios/issues/248

##### SUMMARY
fix error when comparing two vlan using string instead of the int value

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
l2_interfaces.py _check_for_correct_vlan_range


justjais:
@pierremahot thanks for raising the PR for the bug you observed, PR changes LGTM but before we can merge the PR can you add changelog and either integration or unit test for the scenario.

pierremahot:
> @pierremahot thanks for raising the PR for the bug you observed, PR changes LGTM but before we can merge the PR can you add changelog and either integration or unit test for the scenario.

Is the fragment changelog is enough ?

justjais (approved):

PR #253, opened 2021-02-22
Support cisco ios 12.2

mcsSteam:
##### SUMMARY
Cisco IOS 12.2 is not officially supported in ansible. 

During testing with IOS12.2 it was recognized that the facts fetching from the config is failing as ios 12.2 does not support 
`show running config | section xyz`. The section command was introduced in cisco ios 15. As a workaround the config fetching part was adapted so ios facts are working also with cisco ios 12.2.

Instead of using the block command a section command was used and the correct config section was filtered. 

##### ISSUE TYPE
- Feature Pull Request

##### ADDITIONAL INFORMATION
Additional information about filtering in ios can be found here: https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/fundamentals/configuration/15_sy/fundamentals-15-sy-book/cf-cli-search.html


mcsSteam:
No yet ready

PR #270, opened 2021-03-07
Fix interface VLAN creation if it does not exist

JackScripter:
##### SUMMARY
Create vlan interface if interface is not present.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ios_l3_interfaces

##### ADDITIONAL INFORMATION
Before this change, each interface vlan must exist on the switch. This can be done using the module ios_interfaces to enable the interface and then using the module ios_l3_interfaces to configure ip address.

After this change, you can only use ios_l3_interfaces to create and configure a vlan interface.

### DIFF
--- a/plugins/module_utils/network/ios/config/l3_interfaces/l3_interfaces.py
+++ b/plugins/module_utils/network/ios/config/l3_interfaces/l3_interfaces.py
@@ -177,7 +177,7 @@ class L3_Interfaces(ConfigBase):
                 if each["name"] == interface["name"]:
                     break
             else:
-                if "." in interface["name"]:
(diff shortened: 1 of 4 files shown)


Andersson007:
@JackScripter hi, thanks for the contribution!
There are linter issues reported (with no details however).
Could you please try to run:
```
tox -eblack plugins/module_utils/network/ios/config/l3_interfaces/l3_interfaces.py
```

Andersson007:
recheck

KB-perByte:
Hi @JackScripter, we are considering an l3_interface implementation with the latest Resource Module implementation. 
We would address this issue in the fresh implementation. Thank you for your contribution.

Issue #111, opened 2020-07-17
Idempotence issue in ios_logging with host and facility


FloLaco:
##### SUMMARY
If we want to add an host and set the facility, the idempotence is not working

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
`ios_logging` module

##### ANSIBLE VERSION

```
ansible 2.9.10
  config file = /home/florian/logm/ansible-test/ansible.cfg
  configured module search path = ['/home/florian/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/florian/.pyenv/versions/3.8.3/envs/ansible/lib/python3.8/site-packages/ansible
  executable location = /home/florian/.pyenv/versions/ansible/bin/ansible
  python version = 3.8.3 (default, Jun  5 2020, 11:41:04) [GCC 4.8.5 20150623 (Red Hat 4.8.5-36)]
```

##### CONFIGURATION

```
HOST_KEY_CHECKING(/home/user/logm/ansible-test/ansible.cfg) = False
```

##### OS / ENVIRONMENT

```
Switch Ports Model              SW Version            SW Image                 
------ ----- -----              ----------            ----------               
*    1 26    WS-C2960+24TC-L    15.0(2)SE6            C2960-LANBASEK9-M        
```

##### STEPS TO REPRODUCE

```yaml
$ cat configure_logging.yaml
---
- name: Get logging facts
  hosts: switch
  gather_facts: no
  tasks:

  - name: Configure logging
    ios_logging:
      dest: host
      name: 10.55.240.143
      facility: local6
      state: present
    register: output

  - name: Configure trap logging
    ios_logging:
      dest: trap
      level: notifications
      state: present
    register: output2

  - name: Show output
    debug: var=output

  - name: Show output
    debug: var=output2
```

##### EXPECTED RESULTS

Should not have any change after the first execution. Here is a example where we are running twice the same playbook.

##### ACTUAL RESULTS

```paste below
$ ansible-playbook -i hosts configure_logging.yaml -k 
SSH password: 

PLAY [Get logging facts] ****************************************************************************************************************************************************************************************************************

TASK [Configure logging] ****************************************************************************************************************************************************************************************************************
changed: [TB_BHD]

TASK [Configure trap logging] ***********************************************************************************************************************************************************************************************************
ok: [TB_BHD]

TASK [Show output] **********************************************************************************************************************************************************************************************************************
ok: [TB_BHD] => {
[...]

PR #277, opened 2021-03-18
Idempotency issue fix for ios logging

KB-perByte:
##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ios_logging module

##### ADDITIONAL INFORMATION

Before this ios_logging module was failing the idempotent behaviour of Ansible while using facility as a parameter within it
Fix for #111 

```paste below

```

### DIFF
--- a/plugins/modules/ios_logging.py
+++ b/plugins/modules/ios_logging.py
@@ -207,6 +207,31 @@ def validate_size(value, module):
         else:
             return value
 
+def is_command_present(w, have):
+    flag = False
+    combined_have = dict()
+
+    for h in have:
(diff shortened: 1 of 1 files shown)


NilashishC:
@KB-perByte the issue reported in #111 seems like a user error. The task is attempting to add two logging configuration lines in the same task without using aggregate. This will, in most cases, result in idempotency issues and is not the correct approach. The following task should fix this:
```yaml
- ios_logging:
    aggregate:
      - {'dest': 'host', 'name': '10.55.240.143'}
      - {'facility': 'local6'}
    state: present
```

PR #301, opened 2021-04-17
fix example deletion with interface

akira6592:
##### SUMMARY

fix example deletion with interface.

The interface that was supposed to be changed was wrong.

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME

- `cisco.ios.ios_lldp_interfaces`


justjais:
recheck

PR #304, opened 2021-04-21
To fix Cisco IOS no log issue and add ignore txt for 2.12

### LINKED ISSUE #289: Mask sensitive keys in task result
Sensitive keys (including but not limited to the ones detected by validate-modules) should have `no_log: True` so that their values are correctly redacted from the task result.
Also bump the min requirement for netcommon up in galaxy.yml `"ansible.netcommon": ">=2.0.1"`.

justjais:
##### SUMMARY

To fix Cisco IOS no log issue and add ignore txt for 2.12. Fixes #289 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ios

##### ADDITIONAL INFORMATION

```paste below

```


justjais:
recheck

plugins/module_utils/network/ios/facts/bgp_address_family/bgp_address_family.py
```diff
@@ -142,7 +144,9 @@ def populate_facts(self, connection, ansible_facts, data=None):
             )
 
             params = utils.remove_empties(
-                utils.validate_config(self.argument_spec, {"config": objs})
+                bgp_af_parser.validate_config(
```
justjais, line 147: @NilashishC are there any more changes expected to fix the no-log issue from RM side as sanity docker tests for devel and 2.11 with same no-log failures?

plugins/modules/ios_user.py
```diff
@@ -566,7 +566,7 @@ def main():
         password_type=dict(default="secret", choices=["secret", "password"]),
         privilege=dict(type="int"),
         view=dict(aliases=["role"]),
-        sshkey=dict(type="list", elements="str"),
+        sshkey=dict(type="list", elements="str", no_log=True),
```
NilashishC, line 569: This is probably not a sensitive attribute since it's the users public key. In such cases we can set `no_log: False` explicitly so that validate-modules is skipped.

justjais: yea, make sense

NilashishC (commented):
For this to work properly we would need to bump netcommon version in galaxy.yml to >=2.0.1 like https://github.com/ansible-collections/cisco.nxos/pull/278/files#diff-1ade2795a7b410581d689fb0faa358b769c2d189b2d549e8f530b706f09b2fb7R5.

Rest LGTM +1

PR #317, opened 2021-05-03
I guess we aren't testing XE right now?

Qalthos:
##### SUMMARY

`authorize` was removed from `ios_config` a while ago, but is still in the setup tasks when `show version` contains "IOS-XE" 

##### ISSUE TYPE

- Tests Pull Request

##### COMPONENT NAME

prepare_ios_tests


justjais:
@Qalthos yes, we aren’t testing IOS-XE but the changes LGTM, so gating the PR. Thanks

PR #332, opened 2021-05-17
Broken doc link in ios_config module

KB-perByte:
##### SUMMARY

Broken doc link in ios_config module

##### ISSUE TYPE

- Docs Pull Request


docs/cisco.ios.ios_config_module.rst
```diff
@@ -507,7 +507,7 @@ Notes
 
 .. note::
    - Tested against IOS 15.6
-   - Abbreviated commands are NOT idempotent, see L (Network FAQ,../network/user_guide/faq.html#why-do-the-config-modules-always-return-changed-true-with-abbreviated-commands).
+   - Abbreviated commands are NOT idempotent, see `Network FAQ <../network/user_guide/faq.html#why-do-the-config-modules-always-return-changed-true-with-abbreviated-commands>`_.
```
samccann, line 510: Sorry for the delay. Putting this here even though it's closed so you can hopefully find it again.  The correct way to do this is with the[ R() option.](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_documenting.html#linking-and-other-format-macros-within-module-documentation) .  So you have to open  the rst file you want to link to  and find the correct anchor.    Anchors are only visible in the raw rst file (you can't see them in Github unless you click raw).

So in this example, the rst file is at https://github.com/ansible/ansible/blob/devel/docs/docsite/rst/network/user_guide/faq.rst
And if you click raw and scroll down, you'll see the following in the area you want to link to:

```
.. _network_faq_no_abbreviations_with_config:

Why do the ``*_config`` modules always return ``changed=true`` with abbreviated commands?
=========================================================================================
```
The `network_faq_no_abbreviations_with_config` is the anchor (don't include the _ at the beginning or the : at the end.

Here is an example - https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/get_url.py#L21

Issue #206, opened 2020-11-11
email address in MOTD not working.


TmyWu:
##### SUMMARY

Using the cisco.ios.ios_banner module to configure motd on router bugs out while using the '@' symbol in mail.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

cisco.ios.ios_banner
##### ANSIBLE VERSION

```paste below
ansible 2.9.6
  config file = /home/tmy30/Ansible-Infopact-Test1/ansible.cfg
  configured module search path = ['/home/tmy30/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.5 (default, Jul 28 2020, 12:59:40) [GCC 9.3.0]
```

##### CONFIGURATION

```paste below
DEFAULT_HOST_LIST(/home/tmy30/Ansible-Infopact-Test1/ansible.cfg) = ['/home/tmy30/Ansible-Infopact-Test1/inventory']
HOST_KEY_CHECKING(/home/tmy30/Ansible-Infopact-Test1/ansible.cfg) = False
```

##### STEPS TO REPRODUCE

---
```
- hosts: ios_devices
  gather_facts: false
  
  tasks:
    - name: Set MOTD
      cisco.ios.ios_banner:
        banner: motd
        text: |
          *                                                         support@example.com *
          *                                                    https://www.example.com/ *
          *******************************************************************************
        state: present
```
...

```yaml

```

##### ACTUAL RESULTS

```paste below
tmy30@INT-PF1YPCCP:~/Ansible-Infopact-Test1$ ansible-playbook test.yml 

PLAY [ios_devices] **************************************************************************************************************************************************************************************************************

TASK [Set MOTD] *****************************************************************************************************************************************************************************************************************
fatal: [ios_device1]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/bin/python3"}, "changed": false, "msg": "ct_KL_Router(config)#*                                                    https://www.example.com/ *\r\n*                                                    https://www.example.com/ *\r\n^\r\n% Invalid input detected at '^' marker.\r\n\r\nInfopact_KL_Router(config)#"}

PLAY RECAP **********************************************************************************************************************************************************************************************************************
ios_device1                : ok=0    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   
```

Issue #156, opened 2020-09-23
ios_banner seems to have problems with change mode detection and ascii-art banners


lotaris-xx:
##### SUMMARY
If I have ascii-art at the top of my banner, ios_banner always detects a change. 
If I remove the ascii art and start with the blank line above the *** line, the banner works fine.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
ios_banner

##### ANSIBLE VERSION

```
ansible 2.9.7
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.6/site-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.6.8 (default, Dec  5 2019, 15:45:45) [GCC 8.3.1 20191121 (Red Hat 8.3.1-5)]

```

##### CONFIGURATION

```
```

##### OS / ENVIRONMENT

```
Cisco IOS XE Software, Version 16.09.04
Cisco IOS Software [Fuji], Catalyst L3 Switch Software (CAT3K_CAA-UNIVERSALK9-M), Version 16.9.4, RELEASE SOFTWARE (fc2)
Technical Support: http://www.cisco.com/techsupport
Copyright (c) 1986-2019 by Cisco Systems, Inc.
Compiled Thu 22-Aug-19 17:33 by mcpre
```

##### STEPS TO REPRODUCE

Using the following all.yml group_var entry:
```
banner_motd_text: |2
  
       _              _ _     _      _   _      _                      _        _
      / \   _ __  ___(_) |__ | | ___| \ | | ___| |___      _____  _ __| | _____| |
     / _ \ | '_ \/ __| | '_ \| |/ _ \  \| |/ _ \ __\ \ /\ / / _ \| '__| |/ / __| |
    / ___ \| | | \__ \ | |_) | |  __/ |\  |  __/ |_ \ V  V / (_) | |  |   <\__ \_|
   /_/   \_\_| |_|___/_|_.__/|_|\___|_| \_|\___|\__| \_/\_/ \___/|_|  |_|\_\___(_)
  
   ********** WARNING **********
  
   Authorized Personnel Only
  
   This system is the property of the Company, and may be accessed by
   authorized users only.  Unauthorized use is strictly prohibited and may be
   subject to criminal prosecution.  The Company may monitor any activity
   or communication on the system and retrieve any information stored within
[...]

Issue #50, opened 2020-05-14
ios_banner breaks with first line chars


pffs:
##### SUMMARY

If a banner contains characters on the first line before the line break, `ios_banner` crashes.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

`ios_banner`

##### ANSIBLE VERSION

```paste below
ansible 2.9.4
  config file = /home/arantz/change-scripts/ansible/ansible.cfg
  configured module search path = ['/home/arantz/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/arantz/change-scripts/ansible/venv3/lib/python3.6/site-packages/ansible
  executable location = /home/arantz/change-scripts/ansible/venv3/bin/ansible
  python version = 3.6.8 (default, Oct  7 2019, 12:59:55) [GCC 8.3.0]
```

##### CONFIGURATION

```paste below
ANSIBLE_PIPELINING(/home/arantz/change-scripts/ansible/ansible.cfg) = True
DEFAULT_CALLBACK_WHITELIST(/home/arantz/change-scripts/ansible/ansible.cfg) = ['timer', 'json_result']
DEFAULT_FORKS(/home/arantz/change-scripts/ansible/ansible.cfg) = 45
DEFAULT_HOST_LIST(/home/arantz/change-scripts/ansible/ansible.cfg) = ['/home/arantz/change-scripts/ansible/inventories']
DEFAULT_LOCAL_TMP(/home/arantz/change-scripts/ansible/ansible.cfg) = /home/arantz/.ansible/tmp/ansible-local-16723s58fhpxr
DEFAULT_LOG_PATH(/home/arantz/change-scripts/ansible/ansible.cfg) = /home/arantz/change-scripts/ansible/logs/log
HOST_KEY_CHECKING(/home/arantz/change-scripts/ansible/ansible.cfg) = False
INTERPRETER_PYTHON(/home/arantz/change-scripts/ansible/ansible.cfg) = auto_silent
PERSISTENT_COMMAND_TIMEOUT(/home/arantz/change-scripts/ansible/ansible.cfg) = 120
PERSISTENT_CONNECT_TIMEOUT(/home/arantz/change-scripts/ansible/ansible.cfg) = 90
```

##### OS / ENVIRONMENT

Ubuntu 18.04 against Cisco IOS

##### STEPS TO REPRODUCE

```yaml
- ios_banner:
    banner: login
    text: "test"
[...]

PR #342, opened 2021-05-31
Banner fix and modification

KB-perByte:
##### SUMMARY

fixes-
#206 #156 #50

desc-
- Banner module not idempotent
- Fails when while space in the beginning/ back text
- The banner issue with ASCII art
- Introduction of new attribute for delimiter

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ios_banner

### DIFF
--- a/plugins/modules/ios_banner.py
+++ b/plugins/modules/ios_banner.py
@@ -105,14 +105,19 @@ def map_obj_to_commands(updates, module):
     commands = list()
     want, have = updates
     state = module.params["state"]
+    delimiting_char = module.params.get("delimiting_char", "@")
     if state == "absent" and "text" in have.keys() and have["text"]:
         commands.append("no banner %s" % module.params["banner"])
     elif state == "present":
-        if want["text"] and (want["text"] != have.get("text")):
+        if have.get("text") and len(have.get("text")) > 1:
+            haved = have.get("text")[:-1]
+        else:
+            haved = ""
+        if want["text"] and (want["text"] != haved):
             banner_cmd = "banner %s" % module.params["banner"]
-            banner_cmd += " @\n"
+            banner_cmd += " {}\n".format(delimiting_char)
             banner_cmd += want["text"].strip("\n")
-            banner_cmd += "\n@"
+            banner_cmd += "\n{}".format(delimiting_char)
             commands.append(banner_cmd)
     return commands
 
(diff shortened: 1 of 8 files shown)


plugins/modules/ios_banner.py
```diff
@@ -43,6 +43,11 @@
     - incoming
     - slip-ppp
     type: str
+  delimiting_char:
```
justjais, line 46: rename to:
```suggestion
  multiline_delimiter:
```
ref: https://docs.ansible.com/ansible/latest/collections/cisco/ios/ios_config_module.html#parameter-multiline_delimiter

tests/integration/targets/ios_banner/tests/cli/multiple-login-exec.yaml
```diff
@@ -29,12 +29,12 @@
...
     banner: exec
-    text: "this is my login banner\nthat as a multiline\nstring\n"
+    text: "this is my login banner\nthat as a multiline\nstring"
     state: present
-    provider: '{{ cli }}'
+    provider: "{{ cli }}"
 
 - debug:
-    msg: '{{ result }}'
+    msg: "{{ result }}"
```
justjais, line 38: u need to add an integration tests with either `@` or `#` to have a verification at integration level as well

justjais (changes requested):

justjais:
@KB-perByte can you update all of the issues which will be fixed with once this PR gets merged

KB-perByte:
@justjais The known banner issues are tagged already

justjais:
@KB-perByte can u link the issue with #issue_id

plugins/modules/ios_banner.py
KB-perByte: @justjais so for the config it is a multiline delimiter and here is delimiting character, kept it same as in the IOS device itself, for less confusion on its use case

KB-perByte:
recheck

justjais (approved):

plugins/modules/ios_banner.py
```diff
@@ -43,6 +43,11 @@
     - incoming
     - slip-ppp
     type: str
+  multiline_delimiter:
```
rohitthakur2590, line 46: it would be great If we show the use case of `multiline_delimiter` in the example.

rohitthakur2590 (changes requested):

KB-perByte:
recheck

rohitthakur2590 (approved):

PR #343, opened 2021-06-01
To update ios ospfv2 passive_interface implementation

### LINKED ISSUE #336: Unable to config `no passive-interface *` via OSPFv2 RM
##### SUMMARY

For the use-case where the user has applied `passive-interface default` configuration beforehand to the OSPF process and needs to apply `no passive-interface interface1/1` config to the respective OSPF process so that it tells the OSPF to not be passive on the specified interface. Currently with the existing implementation user cannot apply a respective configuration which is the valid and used scenario.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

ios_ospfv2

##### ANSIBLE VERSION

```paste below
2.10+
```

##### CONFIGURATION

```paste below
router ospf 100
 router-id 10.17.255.255
 log-adjacency-changes detail
 passive-interface default
 no passive-interface GigabitEthernet0/1
 network 10.17.254.24 0.0.0.1 area 17
```

##### OS / ENVIRONMENT

macos

justjais:
##### SUMMARY

To update ios ospfv2 passive_interface implementation to suppress/unsupress interface at the same time. Fixes #336 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ios_ospfv2

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/network/ios/argspec/ospfv2/ospfv2.py
+++ b/plugins/module_utils/network/ios/argspec/ospfv2/ospfv2.py
@@ -438,7 +438,8 @@ class Ospfv2Args(object):
                             },
                             "type": "dict",
                         },
-                        "passive_interface": {
+                        "passive_interface": {"type": "str"},
+                        "passive_interfaces": {
                             "options": {
                                 "default": {"type": "bool"},
                                 "interface": {
--- a/plugins/module_utils/network/ios/argspec/ospfv2/ospfv2.py
+++ b/plugins/module_utils/network/ios/argspec/ospfv2/ospfv2.py
@@ -38,7 +38,9 @@ class Ospfv2Args(object):
         "config": {
             "options": {
                 "processes": {
-                    "mutually_exclusive": [("passive_interface", "passive_interfaces")],
+                    "mutually_exclusive": [
+                        ("passive_interface", "passive_interfaces")
+                    ],
                     "elements": "dict",
                     "options": {
                         "address_family": {
--- a/plugins/module_utils/network/ios/argspec/ospfv2/ospfv2.py
+++ b/plugins/module_utils/network/ios/argspec/ospfv2/ospfv2.py
@@ -34,9 +34,6 @@ class Ospfv2Args(object):
     """The arg spec for the ios_ospfv2 module
     """
 
-    def __init__(self, **kwargs):
-        pass
-
     argument_spec = {
         "config": {
             "options": {
@@ -441,7 +438,19 @@ class Ospfv2Args(object):
                             },
                             "type": "dict",
                         },
-                        "passive_interface": {"type": "str"},
+                        "passive_interface": {
+                            "options": {
+                                "default": {"type": "bool"},
+                                "interface": {
+                                    "options": {
+                                        "set_interface": {"type": "bool"},
+                                        "name": {"type": "list"},
+                                    },
+                                    "type": "dict",
+                                },
+                            },
+                            "type": "dict",
+                        },
                         "prefix_suppression": {"type": "bool"},
                         "priority": {"type": "int"},
                         "process_id": {"required": True, "type": "int"},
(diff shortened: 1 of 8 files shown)


plugins/module_utils/network/ios/argspec/ospfv2/ospfv2.py
```diff
@@ -441,7 +438,22 @@ def __init__(self, **kwargs):
...
                         },
-                        "passive_interface": {"type": "str"},
+                        "passive_interface": {
+                            "options": {
+                                "default": {"type": "bool"},
+                                "interface": {
+                                    "options": {
+                                        "set_interface": {"type": "bool"},
+                                        "name": {
+                                            "type": "list",
```
NilashishC, line 448: Does the device support a list of interfaces for this command?

NilashishC: If so, can we have a test case with the same? I'm not entirely sure about the rendering logic in https://github.com/ansible-collections/cisco.ios/pull/343/files#diff-1d26f6676fd263a0e6c15719081998043a2b6aefbe670ad7baced6750528a9f2R539.

justjais: yes, the device supports to un-suppress multiple interface at once, ref: https://github.com/ansible-collections/cisco.ios/pull/343/files#diff-2aafdd2b393ea1646b88f3eec30c692711fa078973fe771a3bee662e50a2e896R9, I can include another `interface`

plugins/module_utils/network/ios/facts/ospfv2/ospfv2.py
```diff
@@ -75,6 +64,17 @@ def populate_facts(self, connection, ansible_facts, data=None):
                 )
 
         for process in current.get("processes", []):
```
NilashishC, line 66: I'm not sure if we should keep on rendering `passive_interface` (the old param) in facts. Since it is deprecated.

plugins/module_utils/network/ios/argspec/ospfv2/ospfv2.py
```diff
@@ -442,6 +439,22 @@ def __init__(self, **kwargs):
                             "type": "dict",
                         },
                         "passive_interface": {"type": "str"},
+                        "passive_interfaces": {
```
NilashishC, line 442: Can we make this `mutually_exclusive` with `passive_interface`?

plugins/modules/ios_ospfv2.py
```diff
@@ -745,6 +745,24 @@
               - Suppress routing updates on an interface (GigabitEthernet A/B)
               - Interface name with respective interface number
             type: str
+          passive_interfaces:
```
NilashishC, line 748: Can we add a deprecation message in the description for `passive_interface`?

plugins/module_utils/network/ios/facts/ospfv2/ospfv2.py
justjais: We shouldn’t do that as `passive_interface` will still be in effect for the next 4 release cycles.

plugins/module_utils/network/ios/argspec/ospfv2/ospfv2.py
justjais: updated

plugins/module_utils/network/ios/facts/ospfv2/ospfv2.py
NilashishC: If we keep it, then round trip would fail because of the `mutually_exclusive` clause since the input would have both.

NilashishC (approved):

justjais:
recheck

justjais:
recheck

justjais:
recheck

justjais:
recheck

PR #411, opened 2021-08-31
fix become for privilege level < 15

KB-perByte:
##### SUMMARY

fix become,  when user privilege < 15 on use of become - privilege should be 15
extra checks added as the prompt checks for `#` that is common for all privilege level.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

terminal/ios.py

##### ADDITIONAL INFORMATION

user privilege
```paste below
an-iosl2-02#show privilege 
Current privilege level is 5
```
before -
```
  tasks:
    - ios_command:
        commands: show privilege
      register: privilege
      become: yes
```
result - 
```
    "stdout_lines": [
        [
            "Current privilege level is 5"
        ]
```
now-
```
    "stdout_lines": [
        [
            "Current privilege level is 15"
        ]
```

### DIFF
--- a/plugins/terminal/ios.py
+++ b/plugins/terminal/ios.py
@@ -65,18 +65,18 @@ class TerminalModule(TerminalBase):
             result = self._exec_cli_command(
                 to_bytes(json.dumps(cmd), errors="surrogate_or_strict")
             )
-
-            prompt = self.privilege_level_re.match(result)
-            if not prompt:
-                raise AnsibleConnectionFailure(
-                    "unable to check privilege level [%s]" % result
-                )
-
-            return int(prompt.group(1))
         except AnsibleConnectionFailure as e:
             raise AnsibleConnectionFailure(
                 "unable to fetch privilege, with error: %s" % (e.message)
             )
+
+        prompt = self.privilege_level_re.match(result)
+        if not prompt:
+            raise AnsibleConnectionFailure(
+                "unable to check privilege level [%s]" % result
+            )
+
+        return int(prompt.group(1))
 
     def on_open_shell(self):
         try:
--- a/plugins/terminal/ios.py
+++ b/plugins/terminal/ios.py
@@ -37,6 +37,8 @@ class TerminalModule(TerminalBase):
         re.compile(br"[\r\n]?[\w\+\-\.:\/\[\]]+(?:\([^\)]+\)){0,3}(?:[>#]) ?$")
     ]
 
+    privilege_level_re = re.compile(r"Current privilege level is (\d+)$")
+
     terminal_stderr_re = [
         re.compile(br"% ?Error"),
         # re.compile(br"^% \w+", re.M),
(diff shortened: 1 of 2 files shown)


plugins/terminal/ios.py
```diff
@@ -57,6 +59,25 @@ class TerminalModule(TerminalBase):
...
+            if not prompt:
+                raise AnsibleConnectionFailure(
+                    "unable to check privilege level [%s]" % result
+                )
+
+            return int(prompt.group(1))
+        except AnsibleConnectionFailure as e:
+            raise AnsibleConnectionFailure(
+                "unable to fetch privilege, with error: %s" % (e.message)
+            )
```
Qalthos, line 79: I think your exceptions might be overlapping in a way you don't want.
```suggestion
        except AnsibleConnectionFailure as e:
            raise AnsibleConnectionFailure(
                "unable to fetch privilege, with error: %s" % (e.message)
            )

        prompt = self.privilege_level_re.match(result)
        if not prompt:
            raise AnsibleConnectionFailure(
                "unable to check privilege level [%s]" % result
            )

        return int(prompt.group(1))
```

plugins/terminal/ios.py
```diff
@@ -92,7 +116,11 @@ def on_become(self, passwd=None):
...
             prompt = self._get_prompt()
-            if prompt is None or not prompt.endswith(b"#"):
+            if (
+                prompt is None
+                or not prompt.endswith(b"#")
+                or self.get_privilege_level() != 15
+            ):
                 raise AnsibleConnectionFailure(
                     "failed to elevate privilege to enable mode still at prompt [%s]"
                     % prompt
```
Qalthos, line 126: Do you want to reference the current privilege level here also?

KB-perByte: That might be redundant.

plugins/terminal/ios.py
```diff
@@ -92,10 +116,13 @@ def on_become(self, passwd=None):
...
+            if (
+                prompt is None
+                or not prompt.endswith(b"#")
+                or self.get_privilege_level() != 15
+            ):
                 raise AnsibleConnectionFailure(
-                    "failed to elevate privilege to enable mode still at prompt [%s]"
-                    % prompt
+                    "failed to elevate privilege to enable mode still at prompt"
                 )
```
Qalthos, line 126: So originally, this error told you why Ansible believed that the elevation failed (i.e., "still at prompt [device%]" or whatever. Now there are a variety of reasons we might think we're not at the right privilege level, and I think we should tell the user why we failed rather than just failing with a message that no longer gives the user any information about why.

Looking closer, we _also_ have the overlapping exception issue here. I would probably also retarget the exceptions so that we are not nesting exceptions unreasonably deeply. GitHub won't let me put a suggestion here, but basically turn lines 118 through 132 to:

```
            prompt = self._get_prompt()
            privilege_level = self.get_privilege_level()
        except AnsibleConnectionFailure as e:
            prompt = self._get_prompt()
            raise AnsibleConnectionFailure(
                "failed to elevate privilege to enable mode, at prompt [%s] with error: %s"
                % (prompt, e.message)
            )

        if (
            prompt is None
            or not prompt.endswith(b"#")
            or privilege_level != 15
        ):
            raise AnsibleConnectionFailure(
                "failed to elevate privilege to enable mode, still at level [%d] and prompt [%s]"
                % (privilege_level, prompt)
            )
```

Qalthos (approved):

PR #433, opened 2021-10-02
ios_banner - ix for Cases of Multiple Delimitation Character 

KyleFryer-Sage:
##### SUMMARY
Implementation of more dynamic banner regex string matching to address edge cases of multiple 'C' delimitation characters that may be present in the banner declaration within the device's configuration.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ios_banner.py

##### ADDITIONAL INFORMATION
For cases where more than one 'C' delimitation char is present in the device running configuration 'banner <banner_type>' declaration line e.g. '^CCC'.

For example, the case:
```
banner login ^CCC
test
^C
```

BEFORE: This results in the following IndexError, as split() has a static string passed into it, ending with ' ^C\n', causing the split to not function as desired when multiple 'C' delimitation chars present:
```
ansible_collections/cisco/ios/plugins/modules/ios_banner.py\", line 150, in map_config_to_obj
IndexError: list index out of range
```

AFTER: The fix implementation, where we first pull out a string match with a regex to accommodate the presence of multiple 'C' delimiters, then pass this regex string match into the split function:
```
EXEC /bin/sh -c '/usr/bin/python3 .ansible/tmp/ansible-local-2108ui6dkvh_/ansible-tmp-1633187470.7116272-2146-30692676631387/AnsiballZ_ios_banner.py && sleep 0'

changed: [x.x.x.x] => {
    "changed": true,
    "commands": [
        "banner login @\nNEW BANNER\n@"
    ],
    "invocation": {
        "module_args": {
            "banner": "login",
            "multiline_delimiter": "@",
            "provider": null,
            "state": "present",
            "text": "NEW BANNER\n"
        }
    }
}
```


KB-perByte:
Hi @KyleFryer-Sage, Can you please address the following
- the issue or implementation with the actual device in the PR details.
- a changelog fragment describing the fix.
- you need to run `tox -e black .`  and  `tox -elinters .` for fixing the tox failure in CI
- a unit or integration test for the fix you added.

Note - we have a banner Resource module in our [roadmap](https://github.com/ansible/community/wiki/Network:2021-Till-Dec-Roadmap) the issue can/may be addressed in the module itself, for that you need to raise a feature/issue as we would deprecate the existing module on merging of the resource module.

let me know if you need any help. Thanks

KB-perByte:
recheck

PR #484, opened 2022-01-17
Snmp server feature - user to add v3 auth and encryption password

KB-perByte:
##### SUMMARY

Can configure v3 password for `authentication` and `encryption` under `users`
NOTE- _might cause idempotency check issues as the configured command is not visible via running-config._

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ios_snmp_server
##### ADDITIONAL INFORMATION

```paste below

```


plugins/module_utils/network/ios/argspec/snmp_server/snmp_server.py
```diff
@@ -453,7 +476,7 @@ class Snmp_serverArgs(object):  # pylint: disable=R0903
                             "type": "str",
                         },
                         "version_option": {
-                            "choices": ["auth", "access", "encrypted"],
+                            "choices": ["encrypted"],
```
NilashishC, line 479: Wouldn't this break existing playbooks?

NilashishC: And if just one value is valid for this option, maybe this should be `encrypted: True`? But again, that would be a breaking change.

KB-perByte: Resolving this as per discussion.

GomathiselviS (approved):
LGTM!

PR #499, opened 2022-01-28
Fix multiple bgp_address_family issues.

dalrrard:
##### SUMMARY
Add `set` option in `send_community` to allow backwards compatiblity with older configs.
Add `set` option in `redistribute.connected` to allow ospf redistribution.
Fix issue with ipv6 and peer-group neighbor identification.
Add ability to pull `redistribute` information for address families to conform to argspec.
Fix issue with not pulling `local_as` when defined for neighbors.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
ios_bgp_address_family

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/499-fix-bgp-af-issues.yaml
+++ b/changelogs/fragments/499-fix-bgp-af-issues.yaml
@@ -1,0 +1,8 @@
+---
+bugfixes:
+  - "`ios_bgp_address_family` - Fix multiple bgp_address_family issues.
+    Add `set` option in `send_community` to allow backwards compatiblity with older configs.
+    Add `set` option in `redistribute.connected` to allow ospf redistribution.
+    Fix issue with ipv6 and peer-group neighbor identification.
+    Add ability to pull `redistribute` information for address families to conform to argspec.
+    Fix issue with not pulling `local_as` when defined for neighbors."
--- a/docs/cisco.ios.ios_bgp_address_family_module.rst
+++ b/docs/cisco.ios.ios_bgp_address_family_module.rst
@@ -2542,7 +2542,7 @@ Parameters
                     <td class="elbow-placeholder"></td>
                 <td colspan="5">
                     <div class="ansibleOptionAnchor" id="parameter-"></div>
-                    <b>ipv6_adddress</b>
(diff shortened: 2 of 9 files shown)


KB-perByte:
Hey @dalrrard Please add test cases for the changes you have added under unit tests or integration tests, The changes look good to me, you may want to tag the issue that you have worked on.
Let me know if you need any help.
Thanks.

dalrrard:
@KB-perByte I've added some unit tests and passed them for `platform linux -- Python 3.9.6, pytest-6.2.5, py-1.11.0, pluggy-1.0.0`. Let me know if anything else needs to be done.

KB-perByte:
And @dalrrard, as you have mentioned backward compatibility can you specify the IOS version we are targeting here, with the specific config we are considering in the change. NOTE- we support IOS version 15+ where supporting an older version of IOS would be an internal discussion, I can update you on it post your reply.
Thanks.

dalrrard:
These changes were tested against 15.X and 16.X configs. The `neighbor send-community` has been around in that form since 12.X, but these new OS versions accept and use it to preserve backwards compatibility.

https://blog.ipspace.net/2009/05/bgp-basics-bgp-communities-propagation.html

dalrrard:
> Hey @dalrrard changing `ipv6_adddress` -> `ipv6_address` though it is a spelling mistake in the module is recorded by us, but changing it now would be breaking change, the best solution to it is adding `alias` to it as of now, we can address the same during a major release.

I'll revert that part. In my testing, it looked like _only_ IPv4 worked and that IPv6 and peer-groups have been unusable, so my thought was to fix the spelling now before it became a usable option.

dalrrard:
@KB-perByte I've reverted the `ipv6_adddress` spelling change and added the alias.

KB-perByte (approved):

KB-perByte:
recheck

KB-perByte:
recheck

KB-perByte:
recheck

KB-perByte:
recheck

goneri:
recheck

PR #516, opened 2022-02-23
Fix AttributeError on dictionary check

vrelk:
Check if proto_options is actually a dictionary instead of simply checking if it exists.

##### SUMMARY

This fixes the error `AttributeError: 'bool' object has no attribute 'keys'`, which is a result of simply checking if proto_options exists, rather than making sure it is actually a dictionary.

Adding protocol_options to every entry isn't an option as that results in `Unsupported attribute for standard ACL - protocol_options.`

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ios_acls

##### ADDITIONAL INFORMATION

An example task that results in the error would be
```
- name: CISCO | IOS | Deploying ACL template
  cisco.ios.ios_acls:
    state: overridden
    config:
      - afi: ipv4
        acls:
          - name: TEST_ACL
            acl_type: extended
            aces:
              - sequence: 10
                grant: permit
                protocol_options:
                  udp: 'yes'
                source:
                  any: 'yes'
                  port_protocol:
                    eq: 68
                destination:
                  any: 'yes'
                  port_protocol:
                    eq: 69
```


KB-perByte:
Hey @vrelk you might want to add a changelog and a test case to check the change a unit test case should be fine, and if it fixes any open issues then you can tag it with the PR.
Regards

KB-perByte (approved):

goneri:
recheck

vrelk:
Yeah, I'm not sure what changed.  Everything passed on be3ad82 except a missing blank line at the end of the changelog, but after fixing that, all hell broke loose.

KB-perByte:
and about the CI failures, there are some changes on our CI pipeline, I am tracking the PRs to be merged.

vrelk:
Err, yeah. Don't think I did that right.... 146 files changed, but at least it passed. I'm just gonna start over. I will reference this pull request in the new one for reference, and put that one here once I have it.

PR #520, opened 2022-03-01
Update l2_interfaces.py

serverwrangler:
Added interface description as available collected value.

##### SUMMARY

Needed a way to apply configurations to specific interface ports based on description.  This simplified the playbook design to be able to filter interfaces by port description when executing configuration changes.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

ios.l2_interfaces

##### ADDITIONAL INFORMATION

This will enable the ability to filter/change/set the port description on interfaces when interacting with interfaces that are described via the "description" configuration parameter. 

Example playbook using the l2_interfaces module feature request:

```
- name: Configure Switch Interfaces for ISE and MAB
  hosts: all
  gather_facts: false
  connection: network_cli
  vars:
    vlans_to_ignore:
      - 999
      - 995
    port_desc2search: "Room112"

  tasks:
     - name: IOS Facts
       ios_facts:
        gather_subset:
        - 'min'
        gather_network_resources:
        - l2_interfaces
       register: interfaces_avail
    # - debug: var=interfaces_avail

     - name: Change ISE port configurations
       #Only change port if IOS and ACCESS mode found
       cisco.ios.ios_config:
           parents: "interface {{ item.name }}" # Take the  interface from loop(with_items), then execute the lines(commands)
           lines:
              - "authentication event server dead action authorize vlan 998"
              - "authentication event no-response action authorize vlan 998"
              - "authentication order mab"
              - "authentication port-control auto"
              - "mab"
           save_when: changed
       with_items: "{{ ansible_network_resources.l2_interfaces }}"
       when: 
         - ansible_network_os == 'ios'
         - item.mode is defined and item.mode == 'access' 
         - item.access.vlan is defined and item.access.vlan not in vlans_to_ignore
         - item.description is defined and item.description is search(port_desc2search) 
       register: print_output
```

### DIFF
--- a/changelogs/fragments/520-l2_interfaces-file.yaml
+++ b/changelogs/fragments/520-l2_interfaces-file.yaml
@@ -1,4 +1,6 @@
+---
 minor_changes:
   - line 42 of plugins/module_utils/network/ios/argspec/l2_interfaces/l2_interfaces.py - added description variable for interface config spec
   - line 112 of plugins/module_utils/network/ios/facts/l2_interfaces/l2_interfaces.py - description variable for configuration parsing of interface description
   - lines 48-52 of plugins/modules/ios_l2_interfaces.py - added Human readable description of the local interface.
+...
--- a/changelogs/fragments/520-l2_interfaces-file.yaml
+++ b/changelogs/fragments/520-l2_interfaces-file.yaml
@@ -1,0 +1,4 @@
+minor_changes:
+  - line 42 of plugins/module_utils/network/ios/argspec/l2_interfaces/l2_interfaces.py - added description variable for interface config spec
+  - line 112 of plugins/module_utils/network/ios/facts/l2_interfaces/l2_interfaces.py - description variable for configuration parsing of interface description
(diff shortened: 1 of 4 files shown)


serverwrangler (commented):
Closed other pull-request for serverwrangler:patch-2 due to dependency of additional files changed that was causing the build test to fail.

serverwrangler:
@KB-perByte the change to the l2_interfaces may be an edge case but provided a simple way for a playbook to apply configuration changes by filtering on vlan and interface description for which the interface module did not support.  

Thanks

KB-perByte:
Hey @serverwrangler, this feature would not be considered under the module, as the description attribute is already addressed in the ios_interfaces module, and adding it to the l2_interface module is pointless and out of the scope of l2 as a resource.
I understand your use-case, where you are iterating over facts data and descriptions together with other l2 data makes it easier to handle, but there are other recommended ways to do it.
Moreover, we are working on a filter plugin that would help combine facts so that we can have single structured data to be used. You may keep an eye on the utils collection for the above-mentioned filter plugin.
Regards

KB-perByte:
Closing the PR due to inactivity, please re-open with the review addressed.

PR #526, opened 2022-03-09
[snmp_server] views facts collection corrected

KB-perByte:
##### SUMMARY

snmp_server views facts collection corrected

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

snmp_server

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/network/ios/rm_templates/snmp_server.py
+++ b/plugins/module_utils/network/ios/rm_templates/snmp_server.py
@@ -320,7 +320,7 @@ class Snmp_serverTemplate(NetworkTemplate):
                 (\sversion\s(?P<version>1|3|2c))?
                 (\s(?P<version_option>auth|noauth|priv))?
                 (\svrf\s(?P<vrf>\S+))?
-                (\s(?P<community_string>\w+))?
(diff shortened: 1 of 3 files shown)


plugins/module_utils/network/ios/rm_templates/snmp_server.py
```diff
@@ -320,7 +320,7 @@ def __init__(self, lines=None, module=None):
                 (\sversion\s(?P<version>1|3|2c))?
                 (\s(?P<version_option>auth|noauth|priv))?
                 (\svrf\s(?P<vrf>\S+))?
-                (\s(?P<community_string>\w+))?
+                (\s(?P<community_string>[-\w]+))?
```
GomathiselviS, line 323: @KB-perByte Is this the only special character allowed? If not, can we make this regex pattern more generic?

KB-perByte: yes that is the only special character allowed, and have to collect just a word in the scope

GomathiselviS (approved):

PR #530, opened 2022-03-14
[ios_static_routes] to consider config only with routes

KB-perByte:
##### SUMMARY

ios_static_routes to consider config only with routes

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ios_static_routes

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/network/ios/facts/static_routes/static_routes.py
+++ b/plugins/module_utils/network/ios/facts/static_routes/static_routes.py
@@ -51,7 +51,7 @@ class Static_RoutesFacts(object):
 
     def get_static_routes_data(self, connection):
         return connection.get(
-            "sh running-config | section ^ip route|^ipv6 route"
+            "show running-config | include ^ip route .+ |^ipv6 route .+"
         )
 
     def populate_facts(self, connection, ansible_facts, data=None):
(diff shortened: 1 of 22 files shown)


NilashishC (approved):

KB-perByte:
@NilashishC adding back the facts filtering code as ios 15.6(2)T apparently has a bug that considers route and router under the same command, which in case of the customer is probably using an updated version of IOS, I checked it with 16.03.05 the show command works fine.
example - 
IOS 15.6(2)T
```
an-ios-01#sh running-config | section ^ip route|^ipv6 route
ip route vrf ansible_temp_vrf 192.0.2.0 255.255.255.0 192.0.2.1 tag 50 name test_vrf track 150
ipv6 route 2001:DB8:0:3::/64 2001:DB8:0:3::2 tag 105 name test_v6
ipv6 route 2001:DB8:0:3::/64 GigabitEthernet0/2 2001:DB8:0:3::2 tag 105 name merged_v6
ipv6 router eigrp 10
```
IOS 16.03.05
```
test#sh running-config | section ^ip route|^ipv6 route  
ip route vrf ansible_temp_vrf 192.0.2.0 255.255.255.0 192.0.2.1 tag 50 name test_vrf track 150
ipv6 route 2001:DB8:0:3::/64 2001:DB8:0:3::2 tag 105 name test_v6
ipv6 route 2001:DB8:0:3::/64 GigabitEthernet0/2 2001:DB8:0:3::2 tag 105 name merged_v6
```
Regards

NilashishC:
@KB-perByte Can you try `show running-config | include ^ip route .+ |^ipv6 route .+`?

plugins/module_utils/network/ios/facts/static_routes/static_routes.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, module, subspec="config", options="options"):
 
     def get_static_routes_data(self, connection):
         return connection.get(
-            "sh running-config | section ^ip route|ipv6 route"
+            "sh running-config | section ^ip route|^ipv6 route"
```
NilashishC, line 54: Please refrain from using the short form of the commands (`show` instead of `sh`). This makes caching useless when turned on.

PR #562, opened 2022-04-20
[facts] default value for `gather_subset` is changed to `min` instead of `!config` 

KB-perByte:
##### SUMMARY

default value for `gather_subset` is changed to `min` instead of `!config` 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ios_facts

##### ADDITIONAL INFORMATION

```paste below

```


goneri:
recheck

NilashishC (approved):

PR #656, opened 2022-09-20
Add support for the configuration revert feature

bentole:
##### SUMMARY
Add support for "Configuration Rollback Confirmed Change" / "Configuration Revert" feature

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins.cliconf.ios.py

##### ADDITIONAL INFORMATION
The "Configuration Rollback Confirmed Change" feature has become increasingly popular. 

This feature brings the well-known IOS-XR  "commit confirmed" operation to IOS and IOS-XE. 

The basic idea is that changes made to network equipment must be confirmed within a given time. If the timer expires, the configuration is rolled back to a configuration backup automatically created when entering global configuration mode.

To make use of this feature, configuration changes must be followed by a task that confirms the change. This is performed by issuing the command "configure confirm" in privileged exec mode. See example below.

This feature greatly reduces the risk of network downtime caused by mis-configuration which are often exacerbated by using Ansible or other automation tools.

The feature requires "Configuration Versioning" to be enabled on devices. the only requirement besides that is to change the way a user enters global configuration mode from  "configure terminal" to e.g "configure terminal revert timer idle 5". 

Obviously, this feature can also be achived by entering global configuration mode in a separate task, but building support for this natively by documenting and defining the parametes would raise awareness about this somewhat hidden gem and have more people use it

Sample playbook to illustrate usage:
```
tasks:
- name: set up configuration versioning
  cisco.ios.ios_config:
    lines:
      - "path bootflash:"
    parents: archive

- name: shut down loopback0 for no apparent reason
  vars:
    # use configuration revert
    ansible_ios_configure_revert: true
    # revert timer type: true=idle, false=absolute
    ansible_ios_configure_revert_timer_idle: false
    # rollback timer in minutes
    ansible_ios_configure_revert_timer: 1

  cisco.ios.ios_config:
    lines:
      - shutdown
    parents: interface Loopback0

# If shutting down Loopback0 leaves the device unreachable the above task will fail
# (ansible_command_timeout) so the below task will never execute
- name: commit changes
  cisco.ios.ios_command:
    commands:
      - configure confirm

# Wait an unnerving 1 minute for the device(s) to become reachable again!
# (Still better than reload in|at...)
```


bentole:
this needs some more code changes before release

PR #659, opened 2022-09-22
add commit_confirm support

bentole:
##### SUMMARY
Add support for "Configuration Rollback Confirmed Change" / "Configuration Revert"  /  "Commit Confirmed" feature
introduces cliconf attributes 
```
    ansible_ios_commit_confirm_immediate: (bool)
    ansible_ios_commit_confirm_timeout: (int)
```

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins.cliconf

##### ADDITIONAL INFORMATION
The "Configuration Rollback Confirmed Change" feature has become increasingly popular. 

This feature brings the well-known IOS-XR  "commit confirmed" operation to IOS and IOS-XE. 

The basic idea is that changes made to network equipment must be confirmed within a given time. If the timer expires, the configuration is rolled back to a configuration backup automatically created when entering global configuration mode.

This feature greatly reduces the risk of network downtime caused by misconfiguration which is often exacerbated by using Ansible or other automation tools.

Tested on IOS-XE version 17.3 and IOS 15.9

Sample playbook to illustrate usage:
```
---
- name: Perform some very dangerous task(s)
  hosts: all
  gather_facts: false
  vars:
    # use commit_confirm
    ansible_ios_commit_confirm_immediate: true
    # rollback timer in minutes
    ansible_ios_commit_confirm_timeout: 1

  tasks:
  - name: shutdown Loopback0 for no reason whatsoever!
    cisco.ios.ios_config:
      lines:
        - shutdown
      parents: interface Loopback0

  # If shutting down Loopback0 leaves the device(s) unreachable the above task will fail
  # with error: 
  #  Got command timeout error. Rollback timer active. 
  #  Allow approximately 40 seconds for the device to become reachable again
```


bentole:
> Hey @bentole would you please add integration tests to this PR.
> I am testing it on my end and will get back to you.
> Regards

Hi @KB-perByte ! Thanks for following up on this! I'll add some tests shortly. I know there are some flaws with the timer validation I need to fix. I'll write that up as well. There might be other brain farts too that I have missed, but hopefully nothing too serious 😄 Let's see what the tests brings us, heh.

bentole:
Hi @KB-perByte! Added integration tests. I looked at iosxr cliconf tests for inspiration and made it more elaborate. Maybe too elaborate? 😄 The tests require the DUT to become unreachable during the test. For this to succeed the interface used for connectivity must be defined. I added a variable for this called `interface_name` in vars.main.yaml. This probably needs to be adapted according to the devices used for testing. Btw, Is there a more optimal place to define this? My test devices are virtual CSR1000v @ 17.3.x and IOSv @ 15.9 and all tests passed.

Have a nice weekend!

KB-perByte:
@bentole the changes look good, I made some cosmetic changes to the tests.
We cannot have commit_delay as an attribute, I totally get the idea behind the delay as (time.sleep) before `configure confirm` but within that delay we are not doing any task-specific operation.
Taking out the delay one can still achieve the same use case likewise.

```
  vars:
    ansible_ios_commit_confirmed: true
    ansible_ios_commit_confirmed_timeout: 1

  tasks:
    - name: Commit confirmed with timeout
      cisco.ios.ios_hostname:
        state: merged
        config:
          hostname: dummy

    - name: Wait 30 secs
      ansible.builtin.wait_for:
        timeout: 30

    - name: "Confirm the Commit"
      cisco.ios.ios_command:
        commands:
          - configure confirm
```
I would rather get this merged as you have already done all the hard parts, if you agree on the above discussed I would push the agreed changes to your PR itself and get it merged.
Regards

bentole:
Hi @KB-perByte ! Thanks for the QA! I cannot in good conscience simply say "Ok, go ahead!" 🙈 The reason is that your sample play raises a very good question that I must admit I havent thought of.

First, lets define the operations as explicit and implicit commits. What you are attempting is an explicit commit. This PR is bringing implicit commits to the equation. Now, if you run your play verbosely you will notice that things are committed twice. First by the implicit commit enabled by `ansible_ios_commit_confirmed` and then by your explicit commit named `Confirm the Commit`. The problem is that the implicit commit is executed as part of the task!

NOW, this could bring the user in a potentially bad situation if he/she is unaware somehow that `ansible_ios_commit_confirmed` is enabled, believing that if the proverbial hits the fan during the `Wait 30 secs` task, changes will be rolled back.. 😰 

So.. we need to make sure the user is aware that implicit commit operation is enabled when attempting explict commit. I rewrote your play a little to illustrate

```
  tasks:
   - name: "Use explicit commit confirmed with timeout"
     block:
      - name: "Prepare explict commit confirmed"
        # must fail if commit_confirmed enabled. Force user to disable
        # commit_confirmed for task/block
        cisco.ios.ios_command:
          commands:
            - conf t revert timer 1

      - name: "Perform the perilous task of changing the hostname"
        cisco.ios.ios_hostname:
          state: merged
          config:
            # this will be implicitly committed if commit_confirmed enabled
            hostname: dummy

      - name: Wait 30 secs
        ansible.builtin.wait_for:
          timeout: 30

      - name: "Confirm the Commit"
        # too late if commit_confirmed enabled. Verbose play will output
        # %No Rollback Confirmed Change pending
        cisco.ios.ios_command:
          commands:
            - configure confirm
```

I've added [validation](https://github.com/ansible-collections/cisco.ios/pull/659/commits/da9692ca24430e5e27f23fd56b253e1f02506b33) that stops the play if the user attempts the task named `Prepare explict commit confirmed`. I think it would be wise to include this, but I'm all ears for suggestions.

Otherwise, I agree to skip the commit_delay! With the option of going explicit users can operate as always. However, since you don't enter conf t explicitly in your play, you will be victim to the double commit mentioned above. I believe it's kind of hard to mitigate that? Maybe we should add the same [validation](https://github.com/ansible-collections/cisco.ios/pull/659/commits/da9692ca24430e5e27f23fd56b253e1f02506b33) to `configure confirm` as well. Hopefully users test their plays before running them in production, heh.

And btw, the reason why I wrote this PR in the first place is that a customer of mine, a major player in the oil industry, banned Ansible from read-write operations. They had a risk assesement that concluded that users cant be "trusted" to explicitly enter global configuration mode by using conf t revert. Allowing this option to be set globally would force the use of this, hence this PR. But we must still allow users to turn it off for certain tasks. It might violate their company policy, but that's not our problem 😄 

I believe many will find use of this feature and it will also raise awareness that this feature, yet still a bit crude, also exist in IOS(-XE) not only in IOS-XR... I see too many today still using "reload in", oblivios of this option. Well, enough sell points. I'm kind of biased anyway 😆 

Thanks again!

KB-perByte:
hey @bentole, I understand the _implicit_ and _explicit_ commit and you rightly pointed out the challenges that may come with that. And we have been through this discussion internally for quite some time.
So, we do not want time.sleep to restrict the processing of our collection level code as that can raise multiple issues in the long run and then the exact concern is what or how one uses the delay interval to do something productive other than wait for a rollback to happen or wait for the configuration to be confirmed. And the idea always has been that the user should have the capability to have much control over the automation rather than Ansible doing a lot of validations and restricting from its end. ( the validation you added was neat though ).
And upon looking at it closely we have figured out _commit_confirmed_ as the name of the option doesn't really suit our exact use case. We plan to rename things moving forward with the following PR as
```
    ansible_ios_commit_confirmed ->  ios_commit_confirm_immediate (bool)
    ansible_ios_commit_confirmed_timeout ->  ios_commit_confirm_timeout (int)
```
^ this change is also to be applied as an alias for _IOSXR_
now coming to some description about how we plan to implement the same, (mostly the same with slight drift in the implementation) 

if only `ios_commit_confirm_timeout` is defined we fire `configure terminal revert timer { timer_int_val }`  and not send `configure confirm` and rely on the user to _explicitly_ send a _conf confirm_.
and the question raised what if only `ios_commit_confirm_immediate` is defined then too we send the `configure terminal revert timer { timer_default_int_val }` as specified now with a _default_ timer, but with this we also fire `configure confirm`  after all the commands are fired followed by `end`. (implicit)
and for a case, we have both defined we already have the timer to add to the conf t command and we also confirm the config right after the task execution. Basically an implicit commit.

let me know your views on this.
It's fun discussing this :) 
Appreciate your efforts and time.

bentole:
Hi again @KB-perByte ! Using time.sleep with a "prolonged" timer often leaves a bad taste so I understand completely why you don't want to have that in the code :) The solution you have provided with a refactoring of the variables seems to be the perfect answer to our predicament. Props!

Let me know if there's anything you need from me. I'm more than happy to help! And thanks again for the in-depth review of this!

Sorry for being a bit short today. I'm in a day-lasting meeting 😆 

Best Regards!

KB-perByte:
Hey, @bentole I have pushed changes as per our discussion to this PR itself.
Let me know if you have any concerns about any of the changes.
Regards.

bentole:
> Hey, @bentole I have pushed changes as per our discussion to this PR itself.
> Let me know if you have any concerns about any of the changes.
> Regards.

Hi @KB-perByte ! Thanks for the updates! Changes makes sense and the puzzle seems to be complete :) 

I have one question, though. I noticed the following is removed from code

```
if "command timeout triggered" in error_msg and commit_confirmed:
    exp_return = commit_confirmed_timeout * 60 - command_timeout
    error_msg = (
        "Got command timeout error. Rollback timer is active. "
        f"Allow approximately {exp_return} seconds from now "
        "for the device to become reachable again"
    )
```
I added this initially to inform the user that rollback timer is active in the event of command timeout so he/she can just chill. It might prevent hyperventilation and potentially heart attacks 😆 But I've come to understand that it's probably the users own responsibility to be aware that rollback timers are active and to create handlers for it so it's kind of against policy to have this (?)

Nothing wrong with it, just curious. Good to know in case of any future shenanigans.

As always; keep up the good work! :)

KB-perByte:
Heyy! @bentole I stopped and thought long before taking out that chunk of exception handling.
Few things that convinced me to take it out.
1. command timeout can be triggered by many factors and we would be always prone to produce errors in terms of false positives if that happens ( and one such error can be critically misleading ).
2. if in case such errors are of utmost importance that handling should be a part of netcommon. ( arguable )
3. without the delay functionality in place the chances of a user getting a command timeout is quite rare ( bit dubious about it ).
4. _'it's probably the users own responsibility to be aware that rollback timers are active and to create handlers for it'_ you have already mentioned it correctly. 

Thanks for working with me on this one. 
Note - you may connect with us on [slack](https://join.slack.com/t/ansiblenetwork/shared_invite/zt-1lk0wn46o-xdO4wPZgsr2WOquiCP~cIg) it would be great to have more feedback and ideas.

ashwini-mhatre (approved):

PR #668, opened 2022-10-05
[WIP] ios_static_routes: 'summarized' state

marehler:
##### SUMMARY

Customer Requirement:

Our customer would like to add static routes only if there are no existing static summary routes. In other words, not add a route if its destination is a subnet of any existing route. Example: 192.168.50.0/30 shall not be added if there is a static route with destination 192.168.50.0/24.

It requires a subnet check for each route to be added. This could be done in an Ansible playbook using a loop and the ansible.netcommon.ipsubnet module, however each iteration takes ~ 0.5 - 1 sec, so that for a large number of routes it is quite slow. The customer is looking for a better performance.

Proposed Feature:

Add a functionality to check for summary routes to the ios_static_routes module. Introduce a "summarized" state that can be thought of as "merged without subroutes". In cisco.ios/plugins/module_utils/network/ios/config/static_routes/static_routes.py, add a case for the new state to 'set_state()', and add a 'remove_sub_routes()' function to remove sub-routes from the 'want' dictionary. 

We are kindly looking for your feedback and advise.

##### ISSUE TYPE

- New Feature Pull Request

##### COMPONENT NAME

cisco.ios.ios_static_routes

##### EXAMPLE

Existing Config:

```
csr1000v-1#show running-config | include ip route|ipv6 route
ip route 198.51.100.0 255.255.255.0 198.51.101.1
```

Playbook Task:

```
- name: Merge configuration if no summary route exists
  cisco.ios.ios_static_routes:
    config:
    - address_families:
      - afi: ipv4
        routes:
        - dest: 198.51.100.0/30
          next_hops:
          - forward_router_address: 198.51.101.1
        - dest: 198.51.102.0/30
          next_hops:
          - forward_router_address: 198.51.101.1
    state: summarized
```

Resulting Config:

```
csr1000v-1#show running-config | include ip route|ipv6 route
ip route 198.51.100.0 255.255.255.0 198.51.101.1
ip route 198.51.102.0 255.255.255.252 198.51.101.1
```

Note: A route for 198.51.100.0/30 has not been added as one for 198.51.100.0/24 already exists.

#### TESTS

So far I have manually tested the remove_sub_routes() function for the following cases:

- 'want' dictionary with one sub-route and one new route per VRF (default and non-default VRF), IPv4 only
- 'want' dictionary with two VRFs (default and non-default VRF), IPv4 and IPv6
- 'want' dictionary with one sub-route in default VRF, resulting in empty VRF
- 'want' dictionary with one sub-route in both VRFs, resulting in empty dictionary

Second, I have tested the adapted ios_static_routes module in an Ansible playbook with 'state: summarized' applied to a CSR1000v running IOS XE version 16.9.3 as shown in the example above.

I would be happy to add integration and unit tests if the feature was accepted.

### DIFF
--- a/docs/cisco.ios.ios_static_routes_module.rst
+++ b/docs/cisco.ios.ios_static_routes_module.rst
@@ -429,6 +429,7 @@ Parameters
                                     <li>gathered</li>
                                     <li>rendered</li>
                                     <li>parsed</li>
+                                    <li>summarized</li>
                         </ul>
                 </td>
                 <td>
@@ -437,6 +438,7 @@ Parameters
                         <div>The state <em>rendered</em> will transform the configuration in <code>config</code> option to platform specific CLI commands which will be returned in the <em>rendered</em> key within the result. For state <em>rendered</em> active connection to remote host is not required.</div>
                         <div>The state <em>gathered</em> will fetch the running configuration from device and transform it into structured data in the format as per the resource module argspec and the value is returned in the <em>gathered</em> key within the result.</div>
                         <div>The state <em>parsed</em> reads the configuration from <code>running_config</code> option and transforms it into JSON format as per the resource module parameters and the value is returned in the <em>parsed</em> key within the result. The value of <code>running_config</code> option should be the same format as the output of command <em>show running-config | include ip route|ipv6 route</em> executed on device. For state <em>parsed</em> active connection to remote host is not required.</div>
+                        <div>The state <em>summarized</em> will add static routes only if there are no existing static summary routes respectively. The check applies per VRF, but does not consider additional attributes like tag or metric. Besides, it behaves like the <em>merged</em> state.</div>
                 </td>
             </tr>
     </table>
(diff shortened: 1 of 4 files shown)


KB-perByte:
Hey @marehler the set of states allowed per resource module is fixed, and a custom state is not allowed, every configuration use case can be achieved by the action states available and for verification/ validation you may choose the non-action states.
We can discuss more over your use case and figure out a solution, but adding a state specific to a resource module wouldn't be allowed.
Regards.

marehler:
Dear Sagar Paul, understood, thank you for the review and feedback. Alternatively to a custom state, would a custom parameter be allowed? If yes, I could use a "no_sub_routes" parameter instead, only to be used with the "merged" state. 

Besides, you are right, the functionally can be achieved with the existing states and Ansible modules. However the desired check if a candidate route shall be configured at all requires some form of loop over all existing routes. Doing this in a playbook is feasible but slow. Hence my approach was to perform it in Python inside an Ansible module. What solution would you suggest for this use case? Appreciate your guidance.

Regards
Martin

KB-perByte:
Hey @marehler adding it as a parameter is definitely recommended and I appreciate your acceptance. Just a note we might not want to associate a parameter with a specific state, it's the user's responsibility to select the correct states with the correct set of attributes. However, it is still doable to achieve a certain use case.
1. I see how you might want to do some specific actions in python rather than in the playbook to get rid of some complexities, we can allow it if and only if it's not possible by normal playbook tasks with validation.
2. For the parameter to be associated with the merged state only, what shall be the behavior of the playbook if the attribute co-exists with other action states?
Can you brief me about the feature a bit more, I would like to understand it better to help you with your contribution.
Regards

marehler:
Hi @KB-perByte , a "no_sub_routes" parameter would be applicable to the state "merged" only. The behaviour would be similar to the "running_config" parameter that is used only with state "parsed". If you use "running_config" with a different state, the playbook raises a failure.

As said, the functionally itself can be achieved with existing playbook tasks, however this is very slow in case of a larger number of existing static routes. The feature basically introduces a check that compares the given "config" parameter with the existing config and removes routes which are kind of redundant because there is already a summary static route covering the subnet. In other words it modifies the given "config" parameter that in turn will be the input to the existing functionality of the module.

Regards
Martin

KB-perByte:
@marehler let's have conversations [here](https://github.com/ansible-collections/cisco.ios/pull/675).

PR #675, opened 2022-10-18
[WIP] ios_static_routes: 'no_subroutes' parameter

marehler:
##### SUMMARY

This pull requests replaces PR #668. It uses a new parameter instead of a custom state as per review comments.

Customer Requirement:

Our customer would like to add static routes only if there are no existing static summary routes. In other words, not add a route if its destination is a subnet of any existing route. Example: 192.168.50.0/30 shall not be added if there is a static route with destination 192.168.50.0/24.

It requires a subnet check for each route to be added. This could be done in an Ansible playbook using a loop and the ansible.netcommon.ipsubnet module, however each iteration takes ~ 0.5 - 1 sec, so that for a large number of routes it is quite slow. The customer is looking for a better performance.

Proposed Feature:

Add a functionality to check for summary routes to the ios_static_routes module. Introduce a "no_subroutes" parameter that can be used with the "merged" state only. If used with other states, it will be silently ignored. In cisco.ios/plugins/module_utils/network/ios/config/static_routes/static_routes.py, add a case for this option to 'set_state()', and add a 'remove_sub_routes()' function to remove sub-routes from the 'want' dictionary.

##### ISSUE TYPE

- New Feature Pull Request

##### COMPONENT NAME

cisco.ios.ios_static_routes

##### EXAMPLE

Existing Config:

```
csr1000v-1#show running-config | include ip route|ipv6 route
ip route 198.51.100.0 255.255.255.0 198.51.101.1
```

Playbook Task:

```
- name: Merge configuration if no summary route exists
  cisco.ios.ios_static_routes:
    config:
    - address_families:
      - afi: ipv4
        routes:
        - dest: 198.51.100.0/30
          next_hops:
          - forward_router_address: 198.51.101.1
        - dest: 198.51.102.0/30
          next_hops:
          - forward_router_address: 198.51.101.1
    no_subroutes: True
```

Resulting Config:

```
csr1000v-1#show running-config | include ip route|ipv6 route
ip route 198.51.100.0 255.255.255.0 198.51.101.1
ip route 198.51.102.0 255.255.255.252 198.51.101.1
```

Note: A route for 198.51.100.0/30 has not been added as one for 198.51.100.0/24 already exists.

#### TESTS

So far I have manually tested the remove_sub_routes() function for the following cases:

- 'want' dictionary with one sub-route and one new route per VRF (default and non-default VRF), IPv4 only
- 'want' dictionary with two VRFs (default and non-default VRF), IPv4 and IPv6
- 'want' dictionary with one sub-route in default VRF, resulting in empty VRF
- 'want' dictionary with one sub-route in both VRFs, resulting in empty dictionary

Second, I have tested the adapted ios_static_routes module in an Ansible playbook with 'no_subroutes: True' applied to a CSR1000v running IOS XE version 16.9.3 as shown in the example above.

I would be happy to add integration and unit tests if the feature was accepted.

### DIFF
--- a/docs/cisco.ios.ios_static_routes_module.rst
+++ b/docs/cisco.ios.ios_static_routes_module.rst
@@ -411,6 +411,25 @@ Parameters
                         <div>The state <em>parsed</em> reads the configuration from <code>running_config</code> option and transforms it into Ansible structured data as per the resource module&#x27;s argspec and the value is then returned in the <em>parsed</em> key within the result.</div>
                 </td>
             </tr>
+
+           <tr>
+                <td colspan="5">
+                    <div class="ansibleOptionAnchor" id="parameter-"></div>
+                    <b>no_subroutes</b>
+                    <a class="ansibleOptionLink" href="#parameter-" title="Permalink to this option"></a>
+                    <div style="font-size: small">
+                        <span style="color: purple">string</span>
+                    </div>
+                </td>
+                <td>
+                </td>
+                <td>
+                        <div>This option is used only with state <em>merged</em>. For other states, it will be ignored.</div>
+                        <div>The default is False.</div>
+                        <div>If set to True, it will add static routes only if there are no existing static summary routes respectively. The check applies per VRF, but does not consider additional attributes like tag or metric.</div>
+                </td>
+            </tr>
+
             <tr>
                 <td colspan="5">
                     <div class="ansibleOptionAnchor" id="parameter-"></div>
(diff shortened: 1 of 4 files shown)


marehler:
Hi Sagar Paul, Understood, however as said the desired summary route check is applicable only for the merge action. For all other action states, this check simply does not make sense. If parameters shall not be bound to a specific state, this may not be the right approach. Are there any alternative solutions you would suggest for this use case? 

Regarding the example, the line is not there in the running config before the action is performed:

# Before state:
# -------------

# vios#show running-config | include ip route|ipv6 route
# ip route 198.51.100.0 255.255.255.0 198.51.101.1

Regards
Martin

marehler:
We concluded to separate the route summary check from the configuration, create a separate custom module to perform this check, and use the ios_static_routes module as is only for the configuration itself.

Issue #680, opened 2022-10-24
'replaced' does not work cisco.ios.ios_route_maps


netexgb:
##### SUMMARY
I have a route-map called TEST_ANSIBLE_COMMUNITY with only one entry, and I am trying to add another entry using replaced, as I reported but #641 where 'merged' does not work, but it looks like 'replaced' does not work either:

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
cisco.ios.ios_route_maps

##### ANSIBLE VERSION
```
ansible --version
ansible [core 2.13.5]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/gbloise/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/gbloise/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/gbloise/.ansible/collections
  executable location = /home/gbloise/.local/bin/ansible
  python version = 3.8.10 (default, Jun 22 2022, 20:18:18) [GCC 9.4.0]
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION
```
ansible-galaxy collection list cisco.ios

# /home/gbloise/.local/lib/python3.8/site-packages/ansible_collections
Collection Version
---------- -------
cisco.ios  4.0.0

# /usr/local/lib/python3.8/dist-packages/ansible_collections
Collection Version
---------- -------
cisco.ios  4.0.0

# /home/gbloise/.ansible/collections/ansible_collections
Collection Version
---------- -------
cisco.ios  4.0.0
```

##### CONFIGURATION
```
ansible-config dump --only-changed
COLLECTIONS_PATHS(/etc/ansible/ansible.cfg) = ['/home/gbloise/.ansible/collections']
DEPRECATION_WARNINGS(/etc/ansible/ansible.cfg) = True
HOST_KEY_CHECKING(/etc/ansible/ansible.cfg) = True
PARAMIKO_HOST_KEY_AUTO_ADD(/etc/ansible/ansible.cfg) = True
PARAMIKO_LOOK_FOR_KEYS(/etc/ansible/ansible.cfg) = True
RETRY_FILES_ENABLED(/etc/ansible/ansible.cfg) = False
TASK_TIMEOUT(/etc/ansible/ansible.cfg) = 0
```

##### OS / ENVIRONMENT
```
us1896-atlanta3-internet-cs#show ver
[...]

Issue #635, opened 2022-08-09
cisco.ios.ios_route_maps does not allow to set a list of community values


netexgb:
##### SUMMARY
I am trying to set a list of community values. In a Cisco router, the "set community" command allows to enter more than one community value, the cisco.ios.ios_route_maps module does not.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
```
cisco.ios.ios_route_maps
```

##### ANSIBLE VERSION

```
ansible --version
ansible [core 2.13.2]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/gbloise/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/gbloise/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /home/gbloise/.ansible/collections
  executable location = /home/gbloise/.local/bin/ansible
  python version = 3.8.10 (default, Mar 15 2022, 12:22:08) [GCC 9.4.0]
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```
ansible-galaxy collection list cisco.ios
# /home/gbloise/.local/lib/python3.8/site-packages/ansible_collections
Collection Version
---------- -------
cisco.ios  3.3.0

# /home/gbloise/.ansible/collections/ansible_collections
Collection Version
---------- -------
cisco.ios  3.3.0

# /usr/local/lib/python3.8/dist-packages/ansible_collections
Collection Version
---------- -------
cisco.ios  3.3.0
```

##### CONFIGURATION

```
In the documentation for cisco.ios.ios_route_maps (https://github.com/ansible-collections/cisco.ios/blob/main/docs/cisco.ios.ios_route_maps_module.rst) the "set" for "community" is the following:

<html>
<body>
<!--StartFragment-->

<html>
<body>
<!--StartFragment-->

setdictionary |   | Match source-protocol of route
-- | -- | --
  |   |   | aigp_metricdictionary |   | accumulated metric value
  |   |   |   | igp_metricboolean | Choices:noyes | metric value from rib
  |   |   |   | valueinteger |   | manual value
  |   |   | as_pathdictionary |   | Prepend string for a BGP AS-path attribute
  |   |   |   | prependdictionary |   | Prepend to the as-path
  |   |   |   |   | as_numberlist / elements=string |   | AS numberPlease refer vendor documentation for valid values
[...]

PR #692, opened 2022-11-20
[ios_route_maps] community issues fix

bentole:
##### SUMMARY
Fixes following issues

* `set community` operations not idempotent
*  community operations not working with replaced state.
* `set community additive` not working properly. Must make sure it's last in list.
* fixed typos in parser.

These issues are discussed in #680 and #635 

Improves or adds the following

* 32-bit number support. Current implementation assumes `ip bgp new-format` is configured.
* merged state made idempotent if `set community` entries exist on device but not part of playbook

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
cisco.ios.ios_route_maps

##### ADDITIONAL INFORMATION

BGP communities are probably the most important detail of Internet Routing. It's vital that these operations work properly.

I added community handling inside a new if block in `list_type_compare`. This is probably not optimal, but I worry about the consequences my code has on other features so wanted to contain it. Besides, I'm too lazy for proper verification beyond integration tests.

Added support for 32-bit number notation of communities in addition to "new-format". We should assume some have reasons to use number instead of new-format. In addition, from a machines perspective it's
easier to handle and understand numbers instead of that weird x:y notation so communities are always converted to numbers by the module. Fret not, IOS(-XE) will automatically convert it back to new-format if `ip bgp new-format` is configured so no special handling required to convert it back.

Also added so merge operations are idempotent when there are existing entries that is not part of the playbook.
E.g if the router has the following configuration 

`set community 65000:100 6500:200` 

and the playbook looks like this
```
community:
 number: 65000:100
```
The code will detect that `65000:100` is already in the community list so changes are skipped

Finally, made sure `additive` keyword is set last in `set community`. IOS(-XE) fails if not.

Updated integration tests with match and set community operations. These tests fails with a vengance on the current main, but are snug like a BUG in a rug with the code in this PR 😄


KB-perByte:
Hey @bentole, I was planning to do something similar for _route_maps_ the way we did for _bgp_address_family_ .
I see a lot of parts in the module that can be optimized and to fix the open issues.
In that case, I won't recommend merging this PR as most of the work would be overridden.
Let me know what you think about it.
Regards

bentole:
> Hey @bentole, I was planning to do something similar for _route_maps_ the way we did for _bgp_address_family_ .
> I see a lot of parts in the module that can be optimized and to fix the open issues.
> In that case, I won't recommend merging this PR as most of the work would be overridden.
> Let me know what you think about it.
> Regards

Hi @KB-perByte ! That sounds awesome! To be honest I dont really expect or care that my PRs be merged. They'more of a FYI and idea sharing. I'm using this in my private fork until a fix in the main appears. Whether that fix comes by merging my PR or by other means is entirely up to you maintainers 😃 

Do what you gotta do and again, keep up the good work!

KB-perByte (approved):

PR #696, opened 2022-11-30
[acls] bug fixes and features

KB-perByte:
##### SUMMARY
ACL bug fixes and features

- _eq_ commands now are interpreted correctly for comparison.  `20 permit udp any any eq 23`  - > `20 permit udp any any eq telnet`
- for _replaced_ and _overridden_ state negate commands are fired before normal commands.
- added support for `object-group-service` under the protocol attribute

##### ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME

ios_acls

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/network/ios/config/acls/acls.py
+++ b/plugins/module_utils/network/ios/config/acls/acls.py
@@ -127,7 +127,7 @@ class Acls(ResourceModule):
             )  # handle aces
 
             self.commands[begin : len(self.commands)] = rearrange_cmds(
-                self.commands[begin : len(self.commands)]
+                self.commands[begin : len(self.commands)],
             )
 
             if len(self.commands) != begin or (not have and want):
--- a/plugins/module_utils/network/ios/config/acls/acls.py
+++ b/plugins/module_utils/network/ios/config/acls/acls.py
@@ -108,7 +108,10 @@ class Acls(ResourceModule):
             acl_type = wentry["acl_type"] if wentry.get("acl_type") else hentry.get("acl_type")
             begin = len(self.commands)  # to determine the index for acl command
             self._compare_aces(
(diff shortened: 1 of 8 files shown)


ashwini-mhatre (commented):
could you please add some tests

plugins/module_utils/network/ios/config/acls/acls.py
```diff
@@ -101,21 +101,35 @@ def _compare(self, want, have, afi):
...
+            begin = len(self.commands)  # to determine the index for acl command
             self._compare_aces(
                 wentry.pop("aces", {}),
                 hentry.pop("aces", {}),
                 afi,
                 wname,
             )  # handle aces
 
+            self.commands[begin : len(self.commands)] = rearrange_cmds(
+                self.commands[begin : len(self.commands)],
```
Qalthos, line 130: I think this is the solution to the sanity issue.
```suggestion
            self.commands[begin:len(self.commands)] = rearrange_cmds(
                self.commands[begin:len(self.commands)],
```

Qalthos (changes requested):

plugins/module_utils/network/ios/config/acls/acls.py
Qalthos: Ugh yeah, black is the one introducing this change. We will have to do something about that.

Qalthos: Is `len(self.commands)` meant to change over the course of execution? Can we not instead do something like this?
```suggestion
            end = len(self.commands)
            self.commands[begin:end] = rearrange_cmds(
                self.commands[begin:],
```

rohitthakur2590 (approved):

PR #723, opened 2023-01-13
add vrf support to L3 interfaces

### LINKED ISSUE #653: Cisco interface resource module does not return forwarding vrf
##### SUMMARY
Unable to use ios_facts gather resources or legacy facts to get forwarding vrf for an interface

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
cisco.ios.ios_facts

##### ADDITIONAL INFORMATION
Module would allow gathering of vrf configuration on interfaces

```text
interface GigabitEthernet0/0/0
 ip vrf forwarding SAMPLE_VRF
```

bentole:
##### SUMMARY
Adds VRF support to L3 interfaces

- Include VRF info in Facts. Fixes #653 
- VRF Interface assignment
- Optional restoration of removed commands
- Legacy support

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
ios.ios_l3_interfaces
ios.ios_vrf

##### ADDITIONAL INFORMATION

**Background**

Ok, trying to be brief. I was recently tasked by a customer of mine to automate changing vrfs on interfaces on hundres of routers. The interfaces were to keep same addressing and configuration as before the vrf change.

I find there's a lot regarding VRFS that I feel is missing so I updated ios.l3_interfaces in my fork
with features necessary to complete the task. I'm now putting this up here for review and debate in case there's any interest of having some isotope of it merged

The features I've added or modified are:

**FACTS**

As far as I can see, neither ios_vrf or l3_interfaces provides VRF facts. This is kind of surprising since a large
number of routers today, if not most of them, use vrfs. ios.l3_interfaces provides ip adressing information, but not vrf information. 
Addng vrfs to the l3_interfaces facts will make the module much more attractive and useable.

**VRF INTERFACE CONFIGURATION**

My original plan was to be a slacker and not do anything beyond the facts since ios.ios_vrf already has the option to assign vrfs to interfaces. But after some pondering I concluded that it doesn't sit right that interface assignment is handled by ios.ios_vrf. My humble opinion is that ios.ios_vrf should handle vrf definitions and ios.l3_interfaces should handle assignments. This would more correctly reflect how vrfs are configured on routers as well. 

Example:

```
# Not like this. Not to critize though, its very convenient :)
- name: assign interface to vrf
  cisco.ios.ios_vrf:
    name: test_vrf
    interfaces:
      - GigabitEthernet5

# Proper way. IMHO
- name: vrf definition
  cisco.ios.ios_vrf:
    name: test_vrf

- name: apply to interface
  ios_l3_interfaces:
    config:
      - name: GigabitEthernet5
      	vrf: test_vrf
    state: merged
```

**RESTORATION**

Changing vrfs on interfaces removes the assigned ip address and ip addresses used for FHRP (HSRP,GLBP), multicast/igmp-joins etc. I noticed that ios.ios_vrf automatically reapplies/restores the ip/ipv6 address(es) when assigning vrfs to interfaces, BUT this automacy contradicts Ciscos behavior of removing the configuration in the first place. Not removing this configuration could in fact be quite dangerous. At least in theory.

It's very convenient that addresses are restored, but it should be a choice made by the user, due to the potential risk involved. In addition, ios.ios_vrf restores the ip/ipv6 address command, but it does not restore other removed commands like the HSRP address. After some back and forth I ended up implementing a code that follows the following principles for restoration

- Automatically restore configuration that is part of the play/intent. Makes sense!
- Do not restore commands negated by the play. Well, Duh...
- Restore commands only if the restore_commands option
  is set, which defaults to false to "comply" with Cisco behavior. Ha!

This is probably better explained by example:

```
# Before state

interface GigabitEthernet5
 ip address 192.168.0.2 255.255.255.0
 ip address 169.254.1.2 255.255.255.0 secondary
 ipv6 address 2001:db8::1/64
 standby 1 ip 192.168.0.1

# Play

- name: apply to interface
  ios_l3_interfaces:
    restore_commands: false
    config:
      - name: GigabitEthernet5
      	vrf: test_vrf
      	ipv4:
      	  - address: 192.168.0.2/24
      	ipv6:
      	  - address: 2001:db8::1/64
    state: merged

# After state

interface GigabitEthernet5
 vrf forwarding test_vrf
 ip address 192.168.0.2 255.255.255.0
 ipv6 address 2001:db8::1/64

```

The remote_command option is set to false (default), but the primary ip/ipv6 addresses are still restored since they are part of the play. The secondary & hsrp addresses are not restored. Running the same play with restore_commands set to true would restore all commands.

**LEGACY SUPPORT**

There are two ways to configure vrfs. `vrf definition` vs `ip vrf`. The latter, is considered legacy, and does not support ipv6, but there might be alot of peeps out there who still use the legacy way for various reasons. To complicate things further, the legacy way has the possibility of running ipv4 in a vrf and ipv6 globally. Finally, I can't find any documentation at cisco where they mention they plan to deprecate the legacy way so decided to add support for this as well.

Legacy way breaks the symbiosis with ios.ios_vrf since that module uses `vrf definition`. A vrf created using `vrf definition`, MUST use vrf forwarding when applied to an interface. A vrf created using `ip vrf`, MUST use `ip vrf forwarding`. IOS returns an error if the oposite is attempted.

I guess legacy way could be used for migrating to the new syntax, but Cisco already has a command for this called vrf `upgrace-cli`. This command does not merge ipv6 into the vrf though. 

Hmm, I just realised that I might have just argued myself away from implementing this, but it's already done so here's an example using legacy way

```
# Before state

interface GigabitEthernet5
 ip address 192.168.0.2 255.255.255.0
 ipv6 address 2001:db8::1/64
 standby 1 ip 192.168.0.1

# Play

- name: apply to interface
  ios_l3_interfaces:
    restore_commands: false
    config:
      - name: GigabitEthernet5
      	ipv4:
      	  - vrf test_vrf
      	  - address: 192.168.0.2/24
    state: merged

# After state

interface GigabitEthernet5
 ip vrf forwarding test_vrf
 ip address 192.168.0.2 255.255.255.0
 ipv6 address 2001:db8::1/64
```

Pay attention to the ipv6 address. It was not part of the play and restore_commands is set to false, yet it's still present! Since ip vrf forwarding does not support ipv6, the ipv6 address is not removed when the vrf is applied. This interface now operates ipv4 in the test_vrf routing table and ipv6 in global routing table.


plugins/module_utils/network/ios/argspec/l3_interfaces/l3_interfaces.py
```diff
@@ -92,6 +94,7 @@ class L3_interfacesArgs(object):  # pylint: disable=R0903
             },
         },
         "running_config": {"type": "str"},
+        "restore_commands": {"type": "bool", "default": "False"},
```
KB-perByte, line 97: @bentole I know this PR is WIP, but restore_commands would be not accepted as a part of the documentation as Ansible argspec doesn't have any declaration for restore_commands, and adding an ad-hoc attribute exclusively to a module is strictly not recommended! The top-level attributes are to be maintained consistently for sake of an argspec being valid for Ansible to recognize it. I am unsure about the functionality you are trying to achieve. Would share some more details regarding it.
You might see sanity already complaining about it.
Regards

bentole: Hi @KB-perByte ! The CI is happy now. The restore_commands is explained in the restoration paragraph. I do understand the concern about breaking the consistency of top-level attributes. I need some help on this. I first had the option together with the vrf parameter so instead of vrf being a string I had

```
 "vrf": {
        "type": "dict",
        "options": {
            "name": {"type": "str"},
            "restore_commands": {"type": "bool"},
        },
}
```
But it felt wrong because restore_commands is referenced to any router command or config. Or is that ok? On top of it all I feel the restoration shouldnt be in the offical release at all due the complexity it adds to the module. On the other hand I believe it's rather cumbersome, if not impossible, to perform consistent restoration using separate tasks (?) At least to accomodate for all eventualities. Being a bit of a devils advocate here 😄

KB-perByte: @bentole this is interesting, but what I can figure out is restore_command would never fire a specific configuration nor flow into facts when we retrieve it. Moreover, the usecase of it is specific to how the config must be applied in when vrf is present. I was just wondering how can we not have a set of logic in the code if we need to do it. And not have a user-facing attribute that changes the course of the whole module execution when enabled.
Note - Facts that are generated are considered source of truth and an attribute not getting enabled/ disabled when facts are generated would have a good impact on the round tripping functionality that the resource module support.
We shall discuss more on this.
Regards.

NilashishC:
Closing this PR as per the conversation above.

PR #764, opened 2023-02-23
update calculations for memtotal/memfree

### LINKED ISSUE #763: ios_facts: memtotal and memfree unit calculations are incorrect
##### SUMMARY

The memtotal and memfree from ios_facts are not converted correctly. 
The name of the fact indicates the figure is in megabytes and the python code divides by 1024. 
Cisco reports this as bytes natively, so to get to megabytes we need to divide by 1048576 instead. 

Additionally, the `memfree_mb` value is being pulled from the wrong array element. It should be pulling from `int(match[2])` not `int(match[3])` 
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

ios_facts
##### ANSIBLE VERSION

```paste below
 ansible [core 2.14.2],
      config file = /etc/ansible/ansible.cfg,
      configured module search path = ['/home/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules'],
      ansible python module location = /usr/lib/python3.9/site-packages/ansible,
      ansible collection location = /home/runner/.ansible/collections:/usr/share/ansible/collections,
      executable location = /usr/bin/ansible,
      python version = 3.9.13 (main, Nov  9 2022, 13:16:24) [GCC 8.5.0 20210514 (Red Hat 8.5.0-15)] (/usr/bin/python3.9),
      jinja version = 3.1.2,
      libyaml = True,
```

##### COLLECTION VERSION

```paste below
    # /usr/share/ansible/collections/ansible_collections,
    Collection Version,
    ---------- -------,
    cisco.ios  4.3.1  ,
```

##### CONFIGURATION

```paste below
    ANSIBLE_FORCE_COLOR(env: ANSIBLE_FORCE_COLOR) = True
    CONFIG_FILE() = /etc/ansible/ansible.cfg
    DEFAULT_CALLBACK_PLUGIN_PATH(env: ANSIBLE_CALLBACK_PLUGINS) = ['/runner/artifacts/246326/callback']
    DEFAULT_LOAD_CALLBACK_PLUGINS(env: ANSIBLE_LOAD_CALLBACK_PLUGINS) = True
    DEFAULT_STDOUT_CALLBACK(env: ANSIBLE_STDOUT_CALLBACK) = awx_display
    HOST_KEY_CHECKING(env: ANSIBLE_HOST_KEY_CHECKING) = False
    INVENTORY_UNPARSED_IS_FAILED(env: ANSIBLE_INVENTORY_UNPARSED_FAILED) = True
    RETRY_FILES_ENABLED(env: ANSIBLE_RETRY_FILES_ENABLED) = False
```

##### OS / ENVIRONMENT

Tested against multiple devices/versions of IOS:

- cisco ISR4351/K9 @ Cisco IOS XE Software, Version 16.12.05
- Cisco CISCO3945-CHASSIS (revision 1.0) with C3900-SPE250/K9 @ Cisco IOS Software, C3900e Software (C3900e-UNIVERSALK9-M), Version 15.7(3)M6
- cisco CSR1000V (VXE) processor (revision VXE) @ Cisco IOS XE Software, Version 17.03.01a <-- Cisco AO DevNet instance

##### STEPS TO REPRODUCE

```yaml
https://gist.github.com/pcunning-rh/d3de2ff338571d6875088dc7131432dc
```

##### EXPECTED RESULTS

I expect the calculations from Byte to Megabyte to be correct
I expect the net_memfree_mb value to come from the `Free` value in the Cisco native output instead of the `Lowest` column.

##### ACTUAL RESULTS

```paste below
"mem_free_mult_by_1024": "477919748.0",
"mem_free_native_from_facts_module": "466718.50390625",
"mem_total_mult_by_1024": "732743660.0",
"mem_total_native_from_facts_module": "715569.98046875",
IOS Command for 'show memory statistics'
"raw_output": [
"Tracekey : 1#048afd6922af1c093e0c3f7a0905f65c   ",
"                Head    Total(b)     Used(b)     Free(b)   Lowest(b)  Largest(b)",
"Processor  7F67FC95F048   732743660   252949748   479793912   477919748   478464024",
"reserve P  7F67FC95F0A0      102404          92      102312      102312      102312",
" lsmpi_io  7F67FB0441A8     3149400     3148576         824         824         412"
```

pcunning-rh:
- IOS reports in Bytes, extra conversions needed to get from Byte to MB as our labels suggest
- memfree_mb is pulling from the wrong element

##### SUMMARY

Fixes #763 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ios_facts
##### ADDITIONAL INFORMATION

```paste below

```


KB-perByte:
Hey @pcunning-rh Please add unit tests and a changelog on basis of the change you added.

pcunning-rh:
Hi @KB-perByte thanks for the review and suggestions re: changelog/unit test.

I've never written a unit test before. Looking through the tests folder, i see a `fixtures` folder; is this intended to be example output for the unit test to run and validate against? It seems the existing [fixtures/ios_facts_show_memory_statistics](https://github.com/ansible-collections/cisco.ios/blob/main/tests/unit/modules/network/ios/fixtures/ios_facts_show_memory_statistics) is empty which suggests a unit test has never been written for this particular scenario, is that correct?

I suspect I would update that file with example output from an ios device and then create an appropriate unit test which then asserts the new appropriate value(s) against that example output?"

changelogs/fragments/ios_facts_memorycalc.yaml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
```
KB-perByte, line 2: This would be a major change, as it impacts all the user directly with this change.
I am wondering if we can keep both somehow.

Moreover, FYI this PR would be held off until we plan a major release for the collection.
The target is we do major releases twice a year.
Regards

pcunning-rh: @KB-perByte Are you proposing creating a new fact with the proper calculation and deprecating the incorrect fact?

KB-perByte: @pcunning-rh Yes that is the right thing to do. If we don't want to introduce a breaking change and for the older facts when the attributes are present a warning can be raised for the same.
Regards

KB-perByte:
@pcunning-rh We had a discussion internally and as the key claims to be mb and it was not processing output as mb, we would be treating it as a bugfix.
I am working forward to merging this.
Thank you for your contribution.

KB-perByte (approved):

Issue #780, opened 2023-03-15
Concatenation issue when setting OSPF redistribute process id in ios_bgp_global


danialre:
##### SUMMARY
When using the ios_bgp_global module, setting the OSPF process ID in the redistribute section results in a module error "an only concatenate str (not \"int\") to str".

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
ios_bgp_global

##### ANSIBLE VERSION
```
ansible [core 2.14.3]
  config file = None
  configured module search path = ['/Users/me/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/me/Library/Python/3.9/lib/python/site-packages/ansible
  ansible collection location = /Users/me/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/me/Library/Python/3.9/bin/ansible
  python version = 3.9.6 (default, Oct 18 2022, 12:41:40) [Clang 14.0.0 (clang-1400.0.29.202)] (/Library/Developer/CommandLineTools/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION
```
# /Users/me/.ansible/collections/ansible_collections
Collection Version
---------- -------
cisco.ios  4.4.0
```

##### CONFIGURATION
```
CONFIG_FILE() = None
```

##### OS / ENVIRONMENT
Tested on macOS Ventura 13.2, as well as Ubuntu 22.04 with Python 3.10 and ansible core 2.14.1

##### STEPS TO REPRODUCE
```
# simple playbook
---
- hosts: all
  gather_facts: no
  connection: network_cli
  tasks:
  - name: Render BGP config
    cisco.ios.ios_bgp_global:
      state: rendered
      config:
        as_number: 64512
        redistribute:
        - ospf:
            process_id: 64512
```

```
# inventory.yml for above playbook
---
all:
  hosts:
    test:
      ansible_host: 10.10.10.10
      ansible_network_os: cisco.ios.ios
```

##### EXPECTED RESULTS
A successful play with an IOS config that looks something like this:
```
router bgp 64512
 address-family ipv4
  redistribute ospf 64512
```

##### ACTUAL RESULTS
A failed play on the example playbook. Extra verbosity doesn't seem to give any details on the error message.
```
PLAY [all] ********************************************************************************************************************************************************************************************

TASK [Render BGP config] ******************************************************************************************************************************************************************************
[...]

PR #687, opened 2022-11-11
[bgp_address_family] bug fix and optimisation

KB-perByte:
##### SUMMARY

bgp_address_family fix & deprecate wrong attribute types, add missing attributes implementations, correct parsers and improve test coverage for the module.

Fixes - 

- Aliased aggregate_address to aggregate_addresses that supports a list of dict attributes (same as bgp_global).
- Aliased neighbor to neighbors that supports a list of dict attributes (same as bgp_global).
- Aliased network to networks that supports a list of dict attributes (same as bgp_global).
- Deprecate slow_peer with slow_peer_options which supports a dict attribute.
- Deprecate neighbors.address/tag/ipv6_adddress with neighbor_address which enables common attributes for facts rendering (same as bgp_global).
- Deprecate neighbors.route_map with route_maps which supports list of dict attributes.
- Deprecate neighbors.prefix_list with prefix_lists which supports list of dict attributes.
- Deprecate neighbors.password with password_options which allows encryption and password.

Bugfixes -

- Added bmp.server_options.
- neighbors.path_attribute to support float format.
- Fix filter list to render correct config.
- Facts and config code optimized for using rm_templates

Features - 

- Added neighbors.advertise
- Added network.evpn
- Added set attributes under redistribute options
- Added parses and config code for 
-  - aggregate-address
-  - auto-summary
-  - non-implimented attributes under bgp
-  - distribute-lists
- Parsers added for non-implemented attributes

##### ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME

ios_bgp_address_family

PR #783, opened 2023-03-19
bgp_address_family: Fix OSPF Redistribution Issues 

bentole:
##### SUMMARY

Fixes issues when redistributing OSPFv2 and OSPFv3 into BGP

- Fixed issue no commands generated
- Fixed `include-connected` option when redistributing OSPF in IPv6 AFI
- Fixed option and syntax for `E1` and `E2` routes
- Fixed option and syntax for `N1` and `N2` routes
- Fixed syntax and regex issues in OSPFv2 and OSPFv3 parsers
- Fixed idempotency issues
- Added appropriate negation of commands in overridden and replaced state

Related issue #780 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
ios_bgp_address_family


plugins/module_utils/network/ios/argspec/bgp_address_family/bgp_address_family.py
```diff
@@ -668,14 +668,25 @@ class Bgp_address_familyArgs(object):  # pylint: disable=R0903
...
-                                                "type_1": {"type": "bool"},
-                                                "type_2": {"type": "bool"},
+                                                "external": {
+                                                    "type": "dict",
+                                                    "options": {
+                                                        "type_1": {"type": "bool"},
+                                                        "type_2": {"type": "bool"},
+                                                    },
+                                                },
+                                                "nssa_external": {
```
KB-perByte, line 683: @bentole we ideally deprecate the attributes which are not perfect by type and re-introduce a new set of attributes, to address it as a bugfix, or else they become a breaking change that requires a major release.

KB-perByte: You may refer to the PR https://github.com/ansible-collections/cisco.ios/pull/687/files for reference on how to deprecate, incase the tonne of things done in this PR confuse you let me know, I would push changes to this PR directly.
Regards

plugins/module_utils/network/ios/rm_templates/bgp_address_family.py
```diff
@@ -2514,41 +2531,53 @@ def __init__(self, lines=None, module=None):
...
-                (\s(?P<type_1>1))?
-                (\s(?P<type_2>2))?
-                (\s(?P<external>external))?
-                (\s(?P<internal>internal))?
-                (\s(?P<nssa_external>nssa-external))?
+                \s+redistribute\sospfv3\s(?P<process_id>\S+)
                 (\smetric\s(?P<metric>\d+))?
+                (\smatch)?
+                (\s(?P<internal>internal))?
+                (\s(?P<ext_type_1>external\s1))?
```
KB-perByte, line 2538: As said we would add fresh attributes with correct options in argspec, we would never show the deprecated attributes in facts but rather handle the option flowing in from the playbook in the config side code, and just render back the correct attributes back in facts.

KB-perByte (changes requested):

plugins/module_utils/network/ios/argspec/bgp_address_family/bgp_address_family.py
bentole: Hi @KB-perByte ! I have made an attempt, emphasis on attempt, to address the issues you were requesting. Ideally I'd like to keep the `external` and `nssa_external` params just changing their type from `bool` to `dict`, but that's probably not possible until the old ones are removed?  Instead I introduced `externals` and `nssa_externals`. Also made an attempt to rewrite the changelog fragment. Hope I'm closer to compliance 😃

KB-perByte:
I am holding this PR in this release. Sorry for the inconvenience!

bentole:
Hi @KB-perByte ! I need to be spoon-fed 😅 If I understand correctly stuff should be re-arranged to look something like this:

```
redistribute:
  - connected:
      metric: 10
      set: true
  - ospfs:
      - process_id: 123
      - process_id: 124
        route_map: bar
```
?

I was thinking since redistribute is already a list of dicts it's perfectly fine to repeat elements. This play works perfectly fine with all states:

```
redistribute:
  - connected:
      metric: 10
      set: true
  - ospf:
      process_id: 123
  - ospf:
      process_id: 124
      route_map: bar
```

If the latter is not consistent with assertions I should probably just rewrite the integration test to redistribute the two processes in separate tasks? If we are to introduce a parent attribute like `ospfs` we should probably do the same with isis and eigrp and then we are doing/fixing something currently not addressed by this PR. Probably better to just rewrite the test. What say ye, boss? :)

KB-perByte:
@bentole sorry for the delayed reply! 
You are right, it works and the states work fine. On second thought let's get this merged as of now. And I'll log it somewhere that we need to redo redistribution, it has been introduced as a list waaaay back, but it should have been a dict that has list attributes when needed. I remember adding a [turn](https://github.com/ansible-collections/cisco.ios/blob/b9532a5807c6e1d49e06e1d0b46091e80a3240d4/plugins/module_utils/network/ios/config/bgp_address_family/bgp_address_family.py#L345) specifically for having this work.
it should be as simple as as top-level [dict](https://github.com/ansible-collections/cisco.ios/blob/b9532a5807c6e1d49e06e1d0b46091e80a3240d4/plugins/module_utils/network/ios/argspec/bgp_address_family/bgp_address_family.py#L591) and ospf as [list](https://github.com/ansible-collections/cisco.ios/blob/b9532a5807c6e1d49e06e1d0b46091e80a3240d4/plugins/module_utils/network/ios/argspec/bgp_address_family/bgp_address_family.py#L668) with elements as dict. Basically if we make an attribute list and fix all the possible entries it is somewhat pointless (wrong) from the model design standpoint! 
Sorry, I made you confused.
Getting this PR merged.
Thanks! :)

KB-perByte (approved):

PR #800, opened 2023-04-11
[l3_interfaces] fix behaviour of enable attribute when value false

KB-perByte:
##### SUMMARY

fix behaviour of enable attribute when value false

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below
  - name: Update Interface Status
    cisco.ios.ios_l3_interfaces:
      config:
        - name: Vlan50
          ipv4:
            - address: 192.168.1.3/31
              dhcp:
                # enable: false
                hostname: some_string
              pool: some_string
              secondary: true
      state: merged
    check_mode: true
```


plugins/module_utils/network/ios/rm_templates/l3_interfaces.py
```diff
@@ -47,6 +47,18 @@ def ip_tmplt(config_data):
     return cmd
 
 
+def ipv4_dhcp(config_data):
+    _data = config_data.get("ipv4", {}).get("dhcp")
+    if not _data.get("enable", True):
+        return None
```
Qalthos, line 53: Is this really okay as None? I would expect this to be an empty string.

KB-perByte: Yep, empty string would be better

roverflow (approved):

PR #813, opened 2023-04-25
Update cisco.ios.ios_hostname_module.rst

manstis:
##### SUMMARY
Fixed a formatting issue with the `Parameters` table.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
`cisco.ios`

##### ADDITIONAL INFORMATION
I was just browsing Ansible in general and noticed an unsightly formatting issue.

### DIFF
--- a/docs/cisco.ios.ios_hostname_module.rst
+++ b/docs/cisco.ios.ios_hostname_module.rst
@@ -49,8 +49,7 @@ Parameters
                 </td>
             </tr>
                                 <tr>
-                    <td class="elbow-placeholder"></td>
-                <td colspan="1">
+                <td colspan="2">
                     <div class="ansibleOptionAnchor" id="parameter-"></div>
                     <b>hostname</b>
                     <a class="ansibleOptionLink" href="#parameter-" title="Permalink to this option"></a>
--- a/docs/cisco.ios.ios_hostname_module.rst
+++ b/docs/cisco.ios.ios_hostname_module.rst
@@ -49,7 +49,8 @@ Parameters
                 </td>
             </tr>
                                 <tr>
-                <td colspan="2">
+                    <td class="elbow-placeholder"></td>
+                <td colspan="1">
                     <div class="ansibleOptionAnchor" id="parameter-"></div>
                     <b>hostname</b>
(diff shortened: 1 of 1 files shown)


manstis:
Oeeee.. `[pre-commit-ci]` reverted my change.. perhaps I've modified the _target_ file and not the _source_.

I'll check.

manstis:
I modified the _source_ file and that too has been reverted.

I am clearly doing something wrong.

KB-perByte:
@manstis so our documentation is generated from the module's docstring dynamically,i.e the .rsts are dynamically generated
If you want to update anything in the documentation of ios_hostname [here](https://github.com/ansible-collections/cisco.ios/blob/a7cacbb48209a3e9c4ab08434851f85014100a1e/plugins/modules/ios_hostname.py#L17), is the file. But I notice you are fixing something that you noticed in the html content [here](https://docs.ansible.com/ansible/devel/collections/cisco/ios/ios_hostname_module.html), Can you please share some more details on that, for me to help you better with your contribution?
Regards

manstis:
Hi @KB-perByte here's a screenshot demonstrating the issue:

image

https://docs.ansible.com/ansible/devel/collections/cisco/ios/ios_hostname_module.html#parameters

The `hostname` is indented as the generated HTML contains additional `<td class="elbow-placeholder"></td>` element. This leads to the `colspan` on the subsequent column also being, incorrectly, `1` and not `2` as is the case with the other parameters.

Now.... this could be completely correct and it is my lack of knowledge that is at fault.

The python file to which you link shows `hostname` as a `suboption` of `config` which would then explain the indentation.

I'll let you decide whether the file is wrong or simply my knowledge.

Please advise and I can simply close this PR if needed.

KB-perByte:
@manstis Thanks for looking out for such details, much appreciated.
But the documentation is valid in terms of the indentation.
a playbook simply looks like -
```
- name: Replace commands with the provided configuration
  cisco.ios.ios_hostname:
    config:
      hostname: RouterTest
    state: replaced
```
config is a dictionary, and all supported attributes under config are represented likewise.
Regards

manstis:
Closing. Issue was my understanding.

Issue #826, opened 2023-05-17
ios_l3_interfaces is not able to remove primary ipv4 address


matejv:
##### SUMMARY

When you have multiple IPv4 addresses on an interface and you want to remove the one that is primary (promoting one of current secondary addresses to primary), the module does not complete this action successfully but does not report any errors. This happens because Cisco IOS does not allow removing a primary address when there are secondary addresses on the interface.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

cisco.ios.ios_l3_interfaces

##### ANSIBLE VERSION

```paste below
ansible [core 2.12.5]
  config file = /home/matej/Projects/harfa/ansible.cfg
  configured module search path = ['/home/matej/.pyenv/versions/3.10.1/envs/harfa/lib/python3.10/site-packages/napalm_ansible/modules']
  ansible python module location = /home/matej/.pyenv/versions/3.10.1/envs/harfa/lib/python3.10/site-packages/ansible
  ansible collection location = /home/matej/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/matej/.pyenv/versions/harfa/bin/ansible
  python version = 3.10.1 (main, Dec 25 2021, 09:38:16) [GCC 11.2.0]
  jinja version = 3.1.2
  libyaml = True

```

##### COLLECTION VERSION

```paste below
# /home/matej/.ansible/collections/ansible_collections
Collection Version
---------- -------
cisco.ios  4.5.0 
```

##### CONFIGURATION

```paste below
ANSIBLE_NOCOWS(/home/matej/Projects/harfa/ansible.cfg) = True
DEFAULT_ACTION_PLUGIN_PATH(env: ANSIBLE_ACTION_PLUGINS) = ['/home/matej/.pyenv/versions/3.10.1/envs/harfa/lib/python3.10/site-packages/napalm_ansible/plugins/action']
DEFAULT_HOST_LIST(/home/matej/Projects/harfa/ansible.cfg) = ['/home/matej/Projects/harfa/inventories/testing/hosts']
DEFAULT_INVENTORY_PLUGIN_PATH(/home/matej/Projects/harfa/ansible.cfg) = ['/home/matej/Projects/harfa/inventory_plugins']
DEFAULT_MANAGED_STR(/home/matej/Projects/harfa/ansible.cfg) = This file is managed by ansible project anso/harfa. Do not edit by hand.
[...]

PR #827, opened 2023-05-25
ios_l3_interfaces account for secondary when comparing addresses

matejv:
##### SUMMARY

ios_l3_interfaces module has bugs when changing IPv4 addresses from secondary to primary with `state: replaced` (see #826 for details).

Changes here first compare have and want lists first for non-secondary addresses, and then in another loop for secondary ones. This builds `ip addres ...` and `no ip address ...` in correct order so the final state on router matches wanted state. 

Fixes: #826

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ios_l3_interfaces


KB-perByte:
@matejv Thanks for fixing the issue, can you please add unit tests and a changelog for the change?
Regards.

matejv:
@KB-perByte I believe this is ready. Unit tests showed another issue in my changes which I fixed.

PR #839, opened 2023-06-07
Limit Python 3.6 as minimum in tests

KB-perByte:
##### SUMMARY

Limit Python 3.6 as a minimum in tests

### DIFF
--- a/tests/config.yml
+++ b/tests/config.yml
@@ -1,0 +1,2 @@
+modules:
+  python_requires: '>= 3.6'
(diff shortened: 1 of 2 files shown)


tests/config.yml
```diff
@@ -0,0 +1,3 @@
+---
+modules:
+  python_requires: ">= 3.6"
```
alisonlhart, line 3:
```suggestion
  python_requires: ">=3.6"
```

roverflow (approved):

PR #864, opened 2023-06-23
SNMP user error fix

roverflow:
##### SUMMARY
The command `show snmp user` gives stderror - `%SNMP agent not enabled` when the snmp agent is not enabled

This PR aims to fix the above issue but handling the error

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
- `facts/snmp_server.py`

### DIFF
--- a/changelogs/fragments/snmp_user_bug.yml
+++ b/changelogs/fragments/snmp_user_bug.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - "Fixes error handling for snmp user when snmp agent is not enabled"
+  - "ios_snmp_server - Fixes error handling for snmp user when snmp agent is not enabled"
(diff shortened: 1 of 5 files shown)


plugins/module_utils/network/ios/facts/snmp_server/snmp_server.py
```diff
@@ -47,6 +47,8 @@ def get_snmpv3_user_data(self, connection):
         Note: The seperate method is needed because the snmpv3 user data is not returned within the snmp-server config
         """
         _get_snmpv3_user = connection.get("show snmp user")
+        if "%" in _get_snmpv3_user:
```
KB-perByte, line 50: That is an appliance error msg, not a command output. we should not handle it here

changelogs/fragments/snmp_user_bug.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - "Fixes error handling for snmp user when snmp agent is not enabled"
```
KB-perByte, line 3:
```suggestion
  - "ios_snmp_server - Fixes error handling for snmp user when snmp agent is not enabled"
```

KB-perByte (approved):

PR #894, opened 2023-07-31
bug: Fixe an error with the pep8 sanity check

earendilfr:
##### SUMMARY
Currently, the sanity check failed because the pep8 doesn't like the line like `type("test") == str`
So, I have replaced all the reference indicated by the sanity check with the command `isinstance(<obj>,<type>)`

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

Error impact only the sanity check done with #PR

##### ADDITIONAL INFORMATION
You can see by example the error indicated in the check of this PR
https://github.com/ansible-collections/cisco.ios/actions/runs/5715835267/job/15486033975?pr=809


earendilfr:
I have two questions concerning the failed checks:
- `ack`: What did I need to put in the PR label? I have the term bug in the PR title and bugfixe in PR text but it seems not enough
- I have test that doesn't work with the module `cliconf`: did it's a problem for my PR that is not related to this module.

changelogs/fragments/pep8_bugfix.yml
```diff
@@ -0,0 +1,4 @@
+---
+bugfixes:
```
KB-perByte, line 2: these are trivial changes, would you please update the changelog type

earendilfr: I have replaced by `minor_changes`

KB-perByte:
@earendilfr I'll put up a PR to fix the unit test failures.
Thanks

PR #900, opened 2023-08-04
ios_snmp_server - Enhancement of the `snmp-server user` part

earendilfr:
##### SUMMARY
fixe a part of the configuration for user:
the current configuration doesn't take correctly the user part. So, the configuration for user if push all the time because, on the retrieve configuration, we didn't have the password part.

So, I have done the following change:
- I retrieve also the algorythm used for authentication and privacy
- During the compare part, for the wanting configuration, I remove the password to perform the comparaison
- On the configuration to generate, I don't send the auth and/or privacy part if the password is not defined

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
ios_snmp_server


earendilfr:
The PR seems good.
The red test are related to the following PR:
- #898: related to the error in test for cliconf
- #894: related to the sanity check because use of `type(<obj>) == <type>` in place of `isinstance(<obj>,<type>)`

roverflow (approved):

PR #901, opened 2023-08-07
Codecov Workflow update

roverflow:
##### SUMMARY
Codecov test pr


codecov.yml
```diff
@@ -1,7 +1,12 @@
 ---
 codecov:
-  require_ci_to_pass: true
-comment: false
+  require_ci_to_pass: false
```
KB-perByte, line 3: we would need CI to pass, why do we want it to succeed on fail?

Issue #910, opened 2023-08-16
cisco.ios.ios_service - Error related to the private-config-encryption feature


earendilfr:
##### SUMMARY
The feature `private-config-encryption` is indicated in the list of defaults services that should be enable but, on many switches, this service is not available.
So, this part should be conditioned to the fact that you seen the this services in the current configuration...

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
cisco.ios.ios_service 

##### ANSIBLE VERSION
<!--- Paste verbatim output from "ansible --version" between quotes -->
```paste below
ansible [core 2.15.2]
  config file = /opt/ansible/ansible.cfg
  configured module search path = ['/home/dadmrossetam/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/ansible/ansible8.2/lib64/python3.11/site-packages/ansible
  ansible collection location = /opt/ansible/collections
  executable location = /opt/ansible/ansible8.2/bin/ansible
  python version = 3.11.2 (main, Jun 22 2023, 04:35:24) [GCC 8.5.0 20210514 (Red Hat 8.5.0-18)] (/opt/ansible/ansible8.2/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION
<!--- Paste verbatim output from "ansible-galaxy collection list <namespace>.<collection>"  between the quotes
for example: ansible-galaxy collection list community.general
-->
```paste below
ansible-galaxy collection list cisco.ios

# /opt/ansible/collections/ansible_collections
Collection Version
---------- -------
cisco.ios  5.0.0
```

##### CONFIGURATION
<!--- Paste verbatim output from "ansible-config dump --only-changed" between quotes -->
```paste below
ACHE_PLUGIN(/opt/ansible/ansible.cfg) = jsonfile
CACHE_PLUGIN_CONNECTION(/opt/ansible/ansible.cfg) = /tmp/facts_cache
CACHE_PLUGIN_TIMEOUT(/opt/ansible/ansible.cfg) = 7200
COLLECTIONS_PATHS(/opt/ansible/ansible.cfg) = ['/opt/ansible/collections']
CONFIG_FILE() = /opt/ansible/ansible.cfg
DEFAULT_FILTER_PLUGIN_PATH(/opt/ansible/ansible.cfg) = ['/opt/ansible/filter_plugins']
[...]

PR #911, opened 2023-08-16
ios_service - Bug with private-config-encryption

earendilfr:
##### SUMMARY

Fix bug #910 

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

Add a conditionnal where private_config_encryption is set to `True` for default value only if present in haved configution


earendilfr:
HI @KB-perByte :
Just to be sure to correctly understand: did you want that I add some other tests in unit test file or an example in copy of the ticket to show the problem prior and after the change ?

KB-perByte:
The unit test file, so that the parameter is tested.

KB-perByte (approved):

Issue #928, opened 2023-09-18
[ios_ospfv2]: Populate OSPF configuration error invalid input detected


brahmanim:
##### SUMMARY

When running ios_ospfv2 integration tests from Controller with EE container,
We get an unexpected module failure `invalid input detected`.
On task [Populate OSPF configuration](https://github.com/ansible-collections/cisco.ios/blob/56b92e89d0bf2a7c36772279ca6c4fe963859ada/tests/integration/targets/ios_ospfv2/tests/cli/_populate_config.yaml#L44C11-L44C20)
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

ios_ospfv2
##### ANSIBLE VERSION

```paste below
2.15
```

##### COLLECTION VERSION

```paste below
V.5.0.0
```

##### CONFIGURATION

```paste below
None
```

##### STEPS TO REPRODUCE

```
Run integration ios_command test, the following task failed.
See below the actual results.
```

```yaml
 - process_id: 200
          vrf: blue
          domain_id:
            ip_address:
              address: 192.0.3.1
          max_metric:
            router_lsa: true
            on_startup:
              time: 100
          auto_cost:
            reference_bandwidth: 4
          areas:
            - area_id: "10"
              capability: true
          distribute_list:
            acls:
              - name: 10
                direction: out
              - name: 123
                direction: in
```

##### EXPECTED RESULTS

The task will pass with not errors.

##### ACTUAL RESULTS

```paste below
TASK [/usr/share/ansible/collections/ansible_collections/cisco/ios/tests/integration/targets/ios_ospfv2 : Populate OSPF configuration] ***
task path: /usr/share/ansible/collections/ansible_collections/cisco/ios/tests/integration/targets/ios_ospfv2/tests/cli/_populate_config.yaml:2
redirecting (type: action) cisco.ios.ios_ospfv2 to cisco.ios.ios
redirecting (type: action) cisco.ios.ios_ospfv2 to cisco.ios.ios
fatal: [10.0.205.140]: FAILED! => {
    "changed": false,
    "module_stderr": "auto-cost reference-bandwidth 4\\r\\nauto-cost reference-bandwidth 4\\r\\n    ^\\r\\n% Invalid input detected at '^' marker.\\r\\n\\r\\nRouter(config)#",
    "module_stdout": "",
    "msg": "MODULE FAILURE\\nSee stdout/stderr for the exact error"
}
```

PR #939, opened 2023-10-04
Fix ospfv2 integration test issues

roverflow:
##### SUMMARY
Address the issue of encountering an unexpected module failure related to invalid input while attempting to run iOS OSPFv2 integration tests from the Controller with the EE container.

fixes: #928 

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/changelogs/fragments/fix_ospfv2_integration_test.yml
+++ b/changelogs/fragments/fix_ospfv2_integration_test.yml
@@ -1,3 +1,3 @@
 ---
-bugfixes:
+trivial:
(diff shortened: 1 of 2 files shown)


changelogs/fragments/fix_ospfv2_integration_test.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
```
KB-perByte, line 2: this is test fix , trivial should be fine 
```suggestion
trivial:
```

PR #944, opened 2023-10-12
Size param

Vovilla:
The size ping parameter is in the doc example, but it is absent in options. Also size is absent in ping command template.

The error is looking as
fatal: [RT1]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (ios_ping) module: size. Supported parameters include: afi, count, dest, df_bit, egress, ingress, source, state, timeout, vrf."}


changelogs/changelog.yaml
roverflow, line 1: Hey @Vovilla, thanks for the PR, as for the changelog, we generally don't edit this file, what we would usually do is add a yml file in the fragments folder which describes the change. Could you change this

plugins/module_utils/network/ios/rm_templates/ping.py
roverflow, line 1: I can see that there are no changes to the argspec, could you also add this attribute to the argspec file and also i can see that the regex and facts have not been updated accordingly, could you please do that as well.

roverflow (changes requested):

docs/cisco.ios.ios_ping_module.rst
roverflow, line 1: This is automatically generated. Please don't edit this file manually.

roverflow (changes requested):

PR #595, opened 2022-06-07
IOS Ping module

KB-perByte:
##### SUMMARY

ios_ping Module

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request

##### COMPONENT NAME

ios_ping

##### ADDITIONAL INFORMATION

Playbook

```paste below
    - name: Check ping
      cisco.ios.ios_ping:
        dest: 10.0.150.115
        count: 4
```

PR #963, opened 2023-11-07
Added size param to ios_ping

roverflow:
##### SUMMARY
Added the size param to ping module which previously did not exist

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
- `ios_ping.py`


KB-perByte:
Hey @roverflow size is a valid parameter

```
foo#ping ?
  WORD      Ping destination address or hostname
  ancp      ANCP echo
  clns      CLNS echo
  ethernet  Ethernet echo
  ip        IP echo
  ipv6      IPv6 echo
  srb       srb echo
  tag       Tag encapsulated IP echo
  vrf       Select VPN routing instance
  <cr>      <cr>

foo#ping 8.8.8.8 ?
  Extended-data  specify extended data pattern
  data           specify data pattern
  df-bit         enable do not fragment bit in IP header
  dscp           Specify DSCP value in ASCII/Numeric
  egress         Force egress interface bypassing routing
  ingress        LAN source interface for Ingress
  repeat         specify repeat count
  size           specify datagram size
  source         specify source address or name
  timeout        specify timeout interval
  tos            specify type of service value
  validate       validate reply data
  <cr>           <cr>

foo#ping 8.8.8.8 size ?
  <36-18024>  Datagram size
```
We missed the attribute during our re-write, could you please look into it and consider the missing attribute as an implementation?
Regards

KB-perByte:
ref : https://github.com/ansible-collections/cisco.ios/pull/595/files#diff-105f4bb6f132265a3e2f631c29eef96876f4e0a38ff634f9430969f612009191L52
This was missed when ping was re-implemented. We would need to fix it.

KB-perByte:
@roverflow The module documentation needs to change too! :D

NilashishC (approved):

zendritic:
When could we expect a release that incorporates this change? I will have to revert our collection to a release prior to the rewrite to restore functionality to one of our most used playbooks due to this issue.

PR #1014, opened 2024-01-18
Resolve idempotency issue in SNMP Server user attribute handling

roverflow:
##### SUMMARY
The goal of this pull request is to address the issue in snmp server user attribute where the playbook on consecutive runs becomes idempotent, as passwords cannot be compared, this should not be happening, this pr aims to resolve that.

Some more context - [issue](https://github.com/ansible-collections/cisco.nxos/issues/801) 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME
- ios_snmp_server.py


tests/unit/modules/network/ios/test_ios_snmp_server.py
```diff
@@ -1933,10 +1933,13 @@ def test_ios_snmpv3_user_server_overridden(self):
             },
         }
         overridden = [
+            "no snmp-server user replaceUser replaceUser v3 access 22",
+            "snmp-server user replaceUser replaceUser v3 auth md5 replaceUser access 22",
             "no snmp-server user flow mfamily v3 access 27",
             "snmp-server user flow mfamily v3 access 27",
             "no snmp-server user newuser newfamily v1 access 24",
         ]
+
```
NilashishC, line 1942: Can you also add a test case with `state: parsed` that confirms that the missing facts are rendered now with this patch?

roverflow: sure

tests/unit/modules/network/ios/test_ios_snmp_server.py
```diff
@@ -1933,10 +1933,13 @@ def test_ios_snmpv3_user_server_overridden(self):
             },
         }
         overridden = [
+            "no snmp-server user replaceUser replaceUser v3 access 22",
```
NilashishC, line 1936: For my understanding, does the CLI require the existing config line to be negated before applying the new one?

changelogs/fragments/snmp_idempotancy_fix.yml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
+  - ios_snmp_server - fixed facts not rendering user authentication password information in snmp_server.
```
NilashishC, line 3: I'm curious about how just config changes results in a facts fix. Can you please elaborate? 

Also, a more general suggestion, for patches like this, it would help if you add more details in the PR summary including what changed, an example of how it currently behaves, as well as one that show how it would behave with the patch merged. This gives more context on the problem that you're trying to solve and facilitates more efficient reviews. :-)

roverflow: ahh this is a mistake, will change that

tests/unit/modules/network/ios/test_ios_snmp_server.py
NilashishC: Are you saying that we're partially rendering some config lines right now?

tests/unit/modules/network/ios/test_ios_snmp_server.py
roverflow: from my understanding it can be updated directly without negation, but since the function had a generic behavior it added these.

roverflow: should i make it so it directly updates without adding the negation

tests/unit/modules/network/ios/test_ios_snmp_server.py
roverflow: Giving some context from this [issue](https://github.com/ansible-collections/cisco.nxos/issues/801), sorry for not explaining clearly, as the playbook on constitutive runs should not be idempotent, this pr aim to solve that, will update the pr comments to reflect this explaination

tests/unit/modules/network/ios/test_ios_snmp_server.py
NilashishC: @KB-perByte would have a better understanding of this than me. But we should always generate the minimal set of commands required to move the device config from current state to desired state. So if the CLI itself doesn't need the negate command, we should NOT run it.

NilashishC: @KB-perByte would have a better understanding of this than me. But we should always generate the minimal set of commands required to move the device config from current state to desired state. So if the CLI itself doesn't need the negate command, we should NOT run it.

tests/unit/modules/network/ios/test_ios_snmp_server.py
NilashishC: Okay - in that case the PR title and the changelog entry seems misleading (at least to me), if the facts are already getting rendered correctly.

tests/unit/modules/network/ios/test_ios_snmp_server.py
roverflow: aahh, okay! will wait for his review

tests/unit/modules/network/ios/test_ios_snmp_server.py
roverflow: updated it!

PR #1034, opened 2024-02-15
Integration test fix for ios_user and ios_facts

roverflow:
##### SUMMARY
- Replaced the hash which was being used to test type 5 secret.
- Changed loop logic for a task to include loop_control to fix error - The loop variable
'item' is already in use.

##### ISSUE TYPE
- Test fix

### DIFF
--- a/changelogs/fragments/ios_user_int_test.yml
+++ b/changelogs/fragments/ios_user_int_test.yml
@@ -1,3 +1,0 @@
----
-minor_changes:
-  - Fix ios_user module integration test failure - change the type 5 secret hash.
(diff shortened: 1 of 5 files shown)


changelogs/fragments/ios_user_int_test.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
```
KB-perByte, line 2: this is just a test change!

KB-perByte (changes requested):

changelogs/fragments/ios_user_int_test.yml
roverflow: Should i just have no changelog??

KB-perByte (approved):

PR #1035, opened 2024-02-19
Fix ios_user integration test removing testing user

roverflow:
##### SUMMARY
Fix ios_user integration test removing testing user, added it to the list of users not to purge

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/tests/integration/targets/ios_user/tests/cli/basic.yaml
+++ b/tests/integration/targets/ios_user/tests/cli/basic.yaml
@@ -178,7 +178,6 @@
   register: result_purge
   cisco.ios.ios_user:
     aggregate:
-      - name: ciscoios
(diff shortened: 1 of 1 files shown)


tests/integration/targets/ios_user/tests/cli/basic.yaml
```diff
@@ -179,6 +179,7 @@
   cisco.ios.ios_user:
     aggregate:
       - name: ciscoios
```
KB-perByte, line 181: This was a user not created for the CI system, this can be removed!

roverflow: Okay!

KB-perByte (approved):

Issue #1072, opened 2024-05-10
cisco.ios.ios_logging_global module is NOT idempotent, so isn't behaving correctly


JuanBH255:
##### SUMMARY

cisco.ios.ios_logging_global module is NOT idempotent, so isn't behaving correctly.
Every time I rerun the playbook it looks as if it's not already created, and a change is made.

##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

```paste below
ansible [core 2.16.6]
```

##### COLLECTION VERSION

```paste below
cisco.ios  5.3.0
```

##### CONFIGURATION

```paste below
CONFIG_FILE() = /projects/dmz_configuration_automation/ansible.cfg
DEFAULT_GATHERING(/projects/dmz_configuration_automation/ansible.cfg) = explicit
DEFAULT_HOST_LIST(/projects/dmz_configuration_automation/ansible.cfg) = ['/projects/dmz_configuration_automation/inv.yml']
DEFAULT_STDOUT_CALLBACK(/projects/dmz_configuration_automation/ansible.cfg) = debug
DISPLAY_SKIPPED_HOSTS(/projects/dmz_configuration_automation/ansible.cfg) = False
HOST_KEY_CHECKING(/projects/dmz_configuration_automation/ansible.cfg) = False
RETRY_FILES_ENABLED(/projects/dmz_configuration_automation/ansible.cfg) = False
```

##### OS / ENVIRONMENT

Cisco C8500-12X
Cisco IOS XE Software, Version 17.03.04a

##### STEPS TO REPRODUCE

```yaml
    - name: Logging
      cisco.ios.ios_logging_global:
        config:
          buffered:
            severity: debugging
            size: 10000
          hosts:
            - hostname: 10.1.1.2
        state: merged
      tags: logging

```

##### EXPECTED RESULTS

No change
ok=1 changed=0

##### ACTUAL RESULTS

```paste below
TASK [Logging] *********************************************************************************************************************************task path: /projects/dmz_configuration_automation/test/test.yml:26
changed: [npdzart03] => {
    "after": {
        "buffered": {
            "size": 10000
        },
        "hosts": [
            {
                "filtered": true,
                "host": "1.1.1.1"
            },
            {
                "host": "10.1.1.2"
            }
        ],
        "source_interface": [
            {
                "interface": "GigabitEthernet0",
                "vrf": "Mgmt-intf"
            }
        ]
    },
    "before": {
        "buffered": {
[...]

PR #1074, opened 2024-05-14
Added doc changes for logging_global

roverflow:
##### SUMMARY
This PR fixes - #1072 

Update documentation - ios_logging_global for severity attribute within buffered.

##### ISSUE TYPE
- Docs Pull Request

### DIFF
--- a/plugins/modules/ios_logging_global.py
+++ b/plugins/modules/ios_logging_global.py
@@ -40,7 +40,7 @@ options:
           severity: &severity
             description:
               - Logging severity level
-              - Idempotancy is not maintained for this attribute for the value debugging
+              - Idempotency is not maintained for this attribute for the value 'debugging'
                 since this value is not shown in running-config.
             type: str
             choices: &severity_subgroup
--- a/plugins/modules/ios_logging_global.py
+++ b/plugins/modules/ios_logging_global.py
@@ -38,7 +38,9 @@ options:
             description: Logging buffer size
             type: int
           severity: &severity
-            description: Logging severity level
+            description: 
+              - Logging severity level
+              - Idempotancy is not maintained for this attribute as value not shown in running-config
             type: str
             choices: &severity_subgroup
               - alerts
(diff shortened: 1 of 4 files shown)


plugins/modules/ios_logging_global.py
```diff
@@ -38,7 +38,9 @@
             description: Logging buffer size
             type: int
           severity: &severity
-            description: Logging severity level
+            description:
+              - Logging severity level
+              - Idempotancy is not maintained for this attribute as value not shown in running-config
```
KB-perByte, line 43: maybe this behavior is specific to the version we currently test with, let's add integration tests to check this nature, and add a comment there. Incase this fails the test in the future we would know to update the doc and fix the test.

tests/integration/targets/ios_logging_global/tests/cli/merged.yaml
```diff
@@ -57,5 +57,23 @@
...
+
+    - name: Test - [merged] merge console severity with device configuration (idempotent)
+      register: resultdeb
+      cisco.ios.ios_logging_global: *id001
+
+    - name: Test - [merged] assert that debugging value is not idempotent
+      ansible.builtin.assert:
+        that:
+          - resultdeb['changed'] == true
+          - "'logging buffered xml 5099 debugging' in resultdeb['commands']"
```
KB-perByte, line 76: assert the after dict to confirm that it impacts idempotency.

KB-perByte (changes requested):

plugins/modules/ios_logging_global.py
```diff
@@ -38,7 +38,10 @@
             description: Logging buffer size
             type: int
           severity: &severity
-            description: Logging severity level
+            description:
+              - Logging severity level
+              - Idempotancy is not maintained for this attribute for the value debugging
```
NilashishC, line 43:
```suggestion
              - Idempotency is not maintained for this attribute for the value 'debugging'
```

plugins/modules/ios_logging_global.py
```diff
@@ -38,7 +38,10 @@
...
+                as value not shown in running-config
```
NilashishC, line 44:
```suggestion
                since this value is not shown in running-config.
```

KB-perByte (approved):

PR #1095, opened 2024-07-04
Fix some idempotency issues in static_routes and bgp_global

weippertt:
##### SUMMARY
Found some minor issues on static_routes not idempotent when TenGigabitEthernet Interfaces in use and
within the same troubleshooting fixing also ebgp-multihop recognition

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
static_routes
bgp_global

##### ADDITIONAL INFORMATION
Add TenGigabitEthernet to interface regex to also recognize these type of interfaces in static routes.
Typo fixed and variable usage in bgp_global for ebgp-multihop

### DIFF
--- a/changelogs/fragments/pr-1095-fragment.yaml
+++ b/changelogs/fragments/pr-1095-fragment.yaml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - static_routes - add TenGigabitEthernet as valid interface
+  - bgp_global - fix ebgp_multihop recognnition and hop_count settings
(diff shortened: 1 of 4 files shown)


KB-perByte:
@weippertt can you please add a changelog entry and a unit test case to test the changes added for bgp_global.
Let us know if you need any help! 
Regards!

weippertt:
I'm currently looking to understand the correct implementation of a unit test for bgp_global. 

One thing, is the fixture for bgp config currently in use? 
I think no one is loading the fixture right now. So i think i need implement a test function ... something like

`def test_ios_bgp_global_multihop(self):
 ...
`

Or am i completely wrong?

weippertt:
> I'm currently looking to understand the correct implementation of a unit test for bgp_global.
> 
> One thing, is the fixture for bgp config currently in use? I think no one is loading the fixture right now. So i think i need implement a test function ... something like
> 
> `def test_ios_bgp_global_multihop(self): ... `
> 
> Or am i completely wrong? And is there an easy way to run the unit tests locally?

I added something i thought it was ok ;)

weippertt:
@KB-perByte does it look ok, or should i tweak some parts on the PR?

KB-perByte (approved):

PR #1124, opened 2024-10-09
 Fix BGP ASDOT notation handling in ios_bgp_global module

### LINKED ISSUE #1096: asdot does not support under bgp local_as command
##### SUMMARY
"I tried to use the **cisco.ios.ios_facts** module to retrieve comprehensive BGP information, but I encountered an issue when using ASDOT notation with the **local_as** command."
```
fatal: [xxxx]: FAILED! => changed=false 
  msg: 'argument ''number'' is of type <class ''float''> found in ''config -> neighbors -> local_as''. and we were unable to convert to int: <class ''float''> cannot be converted to an int'
```
 
 
 
 this is my config
```
 router bgp xxxx
 bgp asnotation dot
 bgp log-neighbor-changes
 bgp deterministic-med
 bgp graceful-restart
 no bgp default ipv4-unicast
 neighbor x.x.x.x remote-as 500.65083
 neighbor x.x.x.x local-as 501.65083 no-prepend replace-as
```
##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
cisco.ios.ios_facts
      - bgp_global

##### ANSIBLE VERSION

```paste below
ansible [core 2.16.3]
  config file = 
  configured module search path = 
  ansible python module location = /opt/homebrew/lib/python3.10/site-packages/ansible
  ansible collection location =
  executable location = /opt/homebrew/bin/ansible
  python version = 3.10.13 (main, Aug 24 2023, 12:59:26) [Clang 15.0.0 (clang-1500.0.40.1)] (/opt/homebrew/opt/python@3.10/bin/python3.10)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
  - name: cisco.ios
    source: https://galaxy.ansible.com
    version: 7.0.0
```

##### CONFIGURATION

```paste below
 router bgp xxxx
 bgp asnotation dot
 bgp log-neighbor-changes
 bgp deterministic-med
 bgp graceful-restart
 no bgp default ipv4-unicast
 neighbor x.x.x.x remote-as 500.65083
 neighbor x.x.x.x local-as 501.65083 no-prepend replace-as
```

##### OS / ENVIRONMENT

cisco ios (Version 17.09.05)
C9500-24Y4C
##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```paste below

```

Girish5tri:
SUMMARY
This PR fixes an issue with the ios_bgp_global module where it fails to handle BGP Autonomous System numbers in ASDOT notation, specifically for the local_as command. The module was previously trying to convert ASDOT notation numbers (e.g., "500.65083") to integers, which caused errors when parsing BGP configurations

To resolve this, we implement deprecation of the "number" parameter in the local_as configuration. The number parameter(type int) has been deprecated in favor of asn(type string) in the bgp_global.py argspec file. This modification allows the module to correctly process ASDOT notation by treating these numbers as strings instead of integers.

Fixes : https://github.com/ansible-collections/cisco.ios/issues/1096

 ISSUE TYPE
- Bugfix Pull Request

COMPONENT NAME
cisco.ios.bgp_global

ADDITIONAL INFORMATION
The issue occurred when trying to retrieve or configure BGP information with configurations using ASDOT notation in the local_as command. This resulted in the following error:

```
fatal: [xxxx]: FAILED! => changed=false 
msg: 'argument ''number'' is of type <class ''float''> found in ''config -> neighbors -> local_as''. and we were unable to convert to int: <class ''float''> cannot be converted to an int'

```
configuration:

```
router bgp xxxx
 bgp asnotation dot
 bgp log-neighbor-changes
 bgp deterministic-med
 bgp graceful-restart
 no bgp default ipv4-unicast
 neighbor x.x.x.x remote-as 500.65083
 neighbor x.x.x.x local-as 501.65083 no-prepend replace-as
```

To fix this, the changes were made in the bgp_global.py argspec, rm templates and config files.

```
"local_as": {
    "type": "dict",
    "options": {
        "set": {"type": "bool"},
         "number": {
                  "type": "int",
                   "removed_at_date": "2027-01-31",
                   "removed_from_collection": "cisco.ios",
         },
         "asn": {"type": "str"},
```

These change allows the module to accept and correctly process ASDOT notation in the local_as command. unit and integration tests have been added to verify the correct handling of ASDOT notation in this context.


plugins/module_utils/network/ios/argspec/bgp_global/bgp_global.py
```diff
@@ -500,7 +500,7 @@ class Bgp_globalArgs(object):  # pylint: disable=R0903
                             "type": "dict",
                             "options": {
                                 "set": {"type": "bool"},
-                                "number": {"type": "int"},
+                                "number": {"type": "str"},
```
roverflow, line 503: The change looks valid. Could you add an action state as a test and include an integration test as well? We need to ensure that existing playbooks continue to function as expected. Please try passing the value as an integer in the playbook to verify that this change doesn't impact the current way of usage.

roverflow (changes requested):

plugins/module_utils/network/ios/argspec/bgp_global/bgp_global.py
Girish5tri: Sure will do that thanks

Girish5tri: I have updated the unit test and added the required integration test in the new commit. I also tested the playbook by passing the value as an integer, and it worked correctly.

tests/integration/targets/ios_bgp_global/tests/cli/merged.yaml
```diff
@@ -27,6 +28,14 @@
               remote_as: 100
               shutdown:
                 set: false
+            - neighbor_address: 192.0.2.1
+              remote_as: "500.65083"
+              local_as:
+                number: "501.65083"
```
roverflow, line 34: This should be an integer in the test, lets make sure that people using this can continue to use the the same way they were using before.

plugins/module_utils/network/ios/argspec/bgp_global/bgp_global.py
roverflow: can you add a test passing it an int value and one more passing it a float value, We need to ensure that it works,  allowing users to continue using it the same way as before. If existing playbooks work seamlessly after the change, we can avoid a deprecation process. However, if they don’t, we’ll need to find a way to make the changes compatible. Otherwise, we may have to deprecate the current attribute and introduce a new one.

Girish5tri: Done! I've updated it in the new commit. added tests in both the unit and integration tests. One test passes the value as an integer, and another passes the value as a float. All the tests are passing

KB-perByte:
@Girish5tri  The local_as.number parameter is updated from int to string that might impact the end user anywhere they have gathered facts and it is safe to consider it as a user-facing change.
The changes looks good. But we must follow a proper deprecation cycle.

```
local_as:
   number: int <- deprecated - remove at/as of jan 2027
   asn: str
```
  
Recommended changes - 

> facts - never render number always render asn (basically a change in rm_templates, and other required places for facts generation)
> config -  handle the deprecation if local_as.number is given from playbook (want) update the data under asn and pop off number.

Reference - 
https://github.com/ansible-collections/cisco.ios/blob/c0aeb5a55d9cb2c78152d516fa724641a83ddb3b/plugins/module_utils/network/ios/config/bgp_global/bgp_global.py#L445 < handle deprecation


CC @roverflow

roverflow (approved):

plugins/module_utils/network/ios/argspec/bgp_global/bgp_global.py
roverflow, line 503: Might have to deprecate i

roverflow (changes requested):

plugins/action/ios.py
```diff
@@ -47,6 +47,21 @@ def run(self, tmp=None, task_vars=None):
                 % self._play_context.connection,
             }
 
+        # Handle deprecated 'number' parameter
```
roverflow, line 50: This should not be here

KB-perByte:
@AAYUSH2091 any updates on this?

Issue #1111, opened 2024-08-24
ios_acls: operations fail on cat9000 configured with arp ACL


ericpedersen-ca:
##### SUMMARY
Catalyst 9000 switches support ARP ACLs for dynamic ARP inspection. The ios_acls module doesn't recognize these and tries to include the ACEs in the previous ACL, which causes operations to fail.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
ios_acls

##### ANSIBLE VERSION

```paste below
ansible [core 2.17.2]
  config file = /home/eric/src/cisco-integrations/cisco-ansible/ansible.cfg
  configured module search path = ['/home/eric/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/eric/src/cisco-integrations/cisco-ansible/.venv/lib64/python3.10/site-packages/ansible
  ansible collection location = /home/eric/src/ansible
  executable location = /home/eric/src/cisco-integrations/cisco-ansible/.venv/bin/ansible
  python version = 3.10.14 (main, Apr 17 2024, 00:00:00) [GCC 14.0.1 20240411 (Red Hat 14.0.1-0)] (/home/eric/src/cisco-integrations/cisco-ansible/.venv/bin/python3)
  jinja version = 3.1.4
  libyaml = True
```

##### COLLECTION VERSION

```paste below
>  ansible-galaxy collection list cisco.ios

# /home/eric/src/ansible/ansible_collections
Collection Version
---------- -------
cisco.ios  9.0.1

# /home/eric/src/cisco-integrations/cisco-ansible/.venv/lib/python3.10/site-packages/ansible_collections
Collection Version
---------- -------
cisco.ios  8.0.0

# /home/eric/src/cisco-integrations/cisco-ansible/.venv/lib64/python3.10/site-packages/ansible_collections
Collection Version
---------- -------
cisco.ios  8.0.0
```

##### CONFIGURATION

```paste below
COLLECTIONS_PATHS(/home/eric/src/cisco-integrations/cisco-ansible/ansible.cfg) = ['/home/eric/src/ansible']
CONFIG_FILE() = /home/eric/src/cisco-integrations/cisco-ansible/ansible.cfg
DEPRECATION_WARNINGS(/home/eric/src/cisco-integrations/cisco-ansible/ansible.cfg) = False
EDITOR(env: EDITOR) = /usr/bin/nano
[...]

PR #1131, opened 2024-11-10
Fix module failure when ARP ACLs are present on cat9k device 

Girish5tri:
SUMMARY
This PR fixes an issue with the ios_acls module where it fails when ARP ACLs are present on the device. The module was incorrectly parsing ARP ACL entries, causing errors when configuring extended ACLs with protocol options.

Changed the command from:
`show running-config | section access-list`
to:
`show running-config | section ip access-list`

```
    def get_acl_data(self, connection):
        # Removed the show access-list
        # Removed the show running-config | include ip(v6)* access-list|remark
        return connection.get("show running-config | section 'ip[v6]* access-list'")
```

This ensures the module only processes IP ACLs and ignores ARP ACLs.

Fixes: https://github.com/ansible-collections/cisco.ios/issues/1111

ISSUE TYPE
Bugfix Pull Request

COMPONENT NAME
cisco.ios.ios_acls

ADDITIONAL INFORMATION
Original error:

`fatal: [device]: FAILED! => {"changed": false, "msg": "Unsupported attribute for standard ACL - protocol."}`

Configuration that triggered the issue:
# On device:
```
ip access-list standard 10
 10 permit 192.168.1.0 0.0.0.255
arp access-list arp-test
 permit ip any mac any
```

Unit tests added to verify the fix. Integration tests present already cover this scenario.

### DIFF
--- a/tests/unit/modules/network/ios/test_ios_acls.py
+++ b/tests/unit/modules/network/ios/test_ios_acls.py
@@ -2303,3 +2303,59 @@ class TestIosAclsModule(TestIosModule):
             "no ip access-list extended test_acl",
         ]
         self.assertEqual(sorted(result["commands"]), sorted(commands))
+
+    def test_ios_acls_with_arp_acl(self):
+        self.execute_show_command.return_value = dedent(
+            """\
+            ip access-list standard 10
+            10 permit 192.168.1.0 0.0.0.255
+            arp access-list arp-test
+            permit ip any mac any
+            ip access-list extended ext_acl
+            10 permit ip 192.0.2.0 0.0.0.255 192.0.3.0 0.0.0.255
+            """
+        )
+        self.execute_show_command_name.return_value = dedent(
+            """\
+            Standard IP access list 10
+            Extended IP access list ext_acl
+            """
+        )
+
+        set_module_args(
+            dict(
+                config=[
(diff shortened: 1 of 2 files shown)


tests/unit/modules/network/ios/test_ios_acls.py
```diff
@@ -2303,3 +2303,59 @@ def test_ios_acls_overridden_remarks_complex(self):
             "no ip access-list extended test_acl",
         ]
         self.assertEqual(sorted(result["commands"]), sorted(commands))
+
+    def test_ios_acls_with_arp_acl(self):
+        self.execute_show_command.return_value = dedent(
+            """\
+            ip access-list standard 10
+            10 permit 192.168.1.0 0.0.0.255
+            arp access-list arp-test
```
KB-perByte, line 2312: Can we please add a integration test where we add a prepare test to add arp access-list and then gather facts so that we are sure with an appliance too that the facts are handled correctly?

KB-perByte (changes requested):

tests/unit/modules/network/ios/test_ios_acls.py
Girish5tri: Sure will do it thanks

PR #1150, opened 2024-12-31
ADD integration workflow for ios

roverflow:
##### SUMMARY
ADD integration workflow for ios

##### ISSUE TYPE

- CI


tests/integration/labs/single.yaml
```diff
@@ -0,0 +1,218 @@
...
+      pae
+      !
+      !
+      crypto pki trustpoint SLA-TrustPoint
+       enrollment pkcs12
+       revocation-check crl
+       hash sha256
+      !
+      !
+      crypto pki certificate chain SLA-TrustPoint
```
KB-perByte, line 69: Do we need to push this as part of the configuration?
Can we not just add the best needed configuration?

tests/integration/labs/single.yaml
```diff
@@ -0,0 +1,218 @@
...
+      !
+      !
+      license udi pid C8000V sn 9EGHOCL6VW2
+      memory free low-watermark processor 201711
+      diagnostic bootup level minimal
+      !
+      !
+      !
+      username cisco privilege 15 secret 9 $9$RelXff9LopGSJE$Sauwc5bEfmrsC1gd5fjzoy7Pe4KDTqvGCPGHBcP.3R.
+      username ansible privilege 15 secret 9 $9$VslpRow9omMy..$5zvxcxrJqLeDd0qEe.5FANSAeLgz9LLAzMiAjieb/nc
```
KB-perByte, line 107: why do we need 2 users? Is it specific to tests?

tests/integration/labs/single.yaml
```diff
@@ -0,0 +1,218 @@
...
+nodes:
+  - boot_disk_size: null
+    configuration: |2-
+      Building configuration...
+
+      Current configuration : 3276 bytes
+      !
+      ! Last configuration change at 08:17:20 UTC Tue Nov 5 2024
+      !
+      version 17.13
```
KB-perByte, line 24: We are only considering a l3 appliance for the upstream tests?

tests/integration/labs/single.yaml
roverflow: This is required as it is needed for ssh

tests/integration/labs/single.yaml
roverflow: removed cisco user

tests/integration/labs/single.yaml
roverflow: currently yes, the cat9k was having issue when trying to setup with upstream, would be working on that next

KB-perByte (approved):

PR #1138, opened 2024-11-25
fix: route-maps ACL removal in replaced state

AAYUSH2091:
##### SUMMARY
Fixes issue with route-maps where ACLs are unnecessarily removed and readded in replaced state. The fix modifies the list_type_compare method to generate only the necessary removal commands for ACLs that are not in the desired state.

Fixes #1134

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cisco.ios.ios_route_maps

PR #1152, opened 2025-01-09
Fixes facts,vrf_interfaces and fixes logic for route maps acls

roverflow:
##### SUMMARY
Fixes facts,vrf_interfaces tests and fixes logic for route maps acls

##### ISSUE TYPE
- Bugfix Pull Request
- Test Fix

##### ADDITIONAL INFORMATION
The above changes fixes facts and vrf_interfaces test cases, it also fixes the logic for acls in route_maps which was modified in this pr - https://github.com/ansible-collections/cisco.ios/pull/1138/files?diff=split&w=0

The above has been tested against cml

### DIFF
--- a/plugins/module_utils/network/ios/config/route_maps/route_maps.py
+++ b/plugins/module_utils/network/ios/config/route_maps/route_maps.py
@@ -31,6 +31,9 @@ from ansible_collections.cisco.ios.plugins.module_utils.network.ios.rm_templates
     Route_mapsTemplate,
 )
 
+# import debugpy
+# debugpy.listen(3000)
+# debugpy.wait_for_client()
 
 class Route_maps(ResourceModule):
     """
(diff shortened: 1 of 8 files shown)


plugins/module_utils/network/ios/config/route_maps/route_maps.py
```diff
@@ -32,6 +32,11 @@
 )
 
 
+# import debugpy
```
NilashishC, line 35: Please remove this.

tests/integration/targets/ios_vrf_interfaces/vars/main.yaml
```diff
@@ -1,7 +1,8 @@
 ---
 gathered:
   config:
-    - name: "Loopback24"
+    - name: "Loopback888"
```
NilashishC, line 4: Are these being cleaned up later?

roverflow: This by default exists in downstream ci but not in cml hence making sure it exists by adding it!

roverflow: The other loopback999 is being removed

NilashishC: Okay, so that means it's not getting cleaned up. We should address that separately.

tests/integration/targets/ios_vrf_interfaces/tests/common/_populate_config.yaml
```diff
@@ -1,12 +1,60 @@
...
+
+- name: Wait for VRF
+  ansible.builtin.wait_for:
+    timeout: 2
+
+- name: Configure interface loopback 1
+  cisco.ios.ios_config:
+    lines:
+      # This exists by default in downstream but not cml
+      - interface Loopback888 #
```
NilashishC, line 39: There's a # after interface name?

PR #1165, opened 2025-03-07
Fix Logging Trap Information

nickbhasin:
##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


plugins/module_utils/network/ios/facts/logging_global/logging_global.py
```diff
@@ -34,7 +36,15 @@ def __init__(self, module, subspec="config", options="options"):
...
     def get_logging_data(self, connection):
-        return connection.get("show running-config | include logging")
+        data = connection.get("show running-config | include logging")
+        if not data:
+            data = connection.get("show logging | include Trap")
+            match = re.search(r"level (\w+)", data)
+            if match:
+                return "logging trap " + match.group(1)
+            else:
+                return ""
```
rohitthakur2590, line 46: Can we use the network template class to parse this?

plugins/module_utils/network/ios/facts/logging_global/logging_global.py
```diff
@@ -34,7 +36,15 @@ def __init__(self, module, subspec="config", options="options"):
         self.argument_spec = Logging_globalArgs.argument_spec
 
     def get_logging_data(self, connection):
-        return connection.get("show running-config | include logging")
+        data = connection.get("show running-config | include logging")
+        if not data:
```
rohitthakur2590, line 40: So this conditional check won't be suitable when the trap is configured with values as "informational" and there are other attributes configured for logging global resources.

rohitthakur2590 (changes requested):

rohitthakur2590:
@nickbhasin  Could you please add changelog , for reference : https://github.com/ansible-collections/cisco.ios/pull/1159/files#diff-b99d4f1136f8bea547af058b41f17ffeca47c2c9daf7c92c402c3168e9ce2f8a 


Also could you please update unit tests and integration test accordingly?

plugins/module_utils/network/ios/facts/logging_global/logging_global.py
nickbhasin: I have fixed the logic. Now it will first check logging trap is in data or not and then fetch the trap and add it to existing values in data variable.

plugins/module_utils/network/ios/facts/logging_global/logging_global.py
nickbhasin: Are you talking about the rm templates class?

nickbhasin: Fixed.

tests/integration/targets/ios_logging_global/tests/cli/merged.yaml
```diff
@@ -32,7 +32,7 @@
           message_counter:
             - log
             - syslog
-          trap: errors
+          trap: informational
```
rohitthakur2590, line 35: this is fine, but we also need to cover scenarios where the trap is other than informational. may be you could use this with replaced operation

tests/unit/modules/network/ios/test_ios_logging_global.py
```diff
@@ -718,3 +719,20 @@ def test_ios_logging_global_replaced_idempotent(self):
 
         self.maxDiff = None
         self.assertEqual(sorted(result["commands"]), sorted(replaced))
+
+    def mock_connection(self):
+        """Fixture to provide a mock connection object"""
+        mock = MagicMock()
+        return mock
+
+    def test_get_logging_data_logging_trap_informational(self):
```
rohitthakur2590, line 728: so what we need to test here is tap: information being part of gathered and merged operation.

tests/integration/targets/ios_logging_global/tests/cli/merged.yaml
rohitthakur2590, line 35: this is fine, but we also need to cover scenarios where the trap is other than informational. may be you could use this with replaced operation

rohitthakur2590 (changes requested):

tests/integration/targets/ios_logging_global/tests/cli/merged.yaml
nickbhasin: Added trap : error in replaced.yaml.

nickbhasin: added trap : errors in main.yaml for replaced state.

rohitthakur2590 (approved):

rohitthakur2590:
We can merge this, once CI failures are resolved

PR #1181, opened 2025-04-24
Fix(ios_interfaces): Prevent unintended no shutdown by removing enabled default

AAYUSH2091:
##### SUMMARY
Fixes incorrect handling of the `enabled` (shutdown/no shutdown) state within the `cisco.ios.ios_interfaces` module when configuration changes detected.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cisco.ios.ios_interfaces

#####INFO 
This change updates the parser to assume interfaces are enabled (no shutdown) by default unless a shutdown command is explicitly found.
It reflects the default Cisco IOS behavior, where interfaces are administratively up unless shut down.
By setting 'enabled': True in the absence of a shutdown, the parser captures the accurate current state ("have" state).
If shutdown is present, existing logic overrides and sets 'enabled': False.
This ensures accurate comparison with the desired state in playbooks.
Ultimately, it prevents unnecessary no shutdown commands from being sent when not specified by the user.
This fix improves the module's ability to read and act on the real configuration.


plugins/module_utils/network/ios/argspec/interfaces/interfaces.py
```diff
@@ -39,7 +39,7 @@ class InterfacesArgs(object):  # pylint: disable=R0903
             "options": {
                 "name": {"type": "str", "required": True},
                 "description": {"type": "str"},
-                "enabled": {"type": "bool", "default": True},
```
Ruchip16, line 42: shouldnt this be true by default? please revert it, we need this change

KB-perByte (approved):

PR #1192, opened 2025-05-19
Adding_stitching_command_supported_in_route_targets(list of dict)

AAYUSH2091:
##### SUMMARY
Enhances `ios_vrf_address_family` facts gathering for `route-target` configurations. This change ensures the `stitching` attribute is correctly parsed and included for both `export` and `import_config` entries. Furthermore, `route_target.export` and `route_target.import_config` will now consistently be lists of dictionaries, each containing the `rt_value` and its `stitching` status (true/false), providing a more structured and complete representation even for single route targets. Previously, the `stitching` attribute was omitted, and the structure might not have been a list of dictionaries for single RTs.

Additionally, this change resolves integration test failures for:
ios_interfaces
ios_lacp_interfaces
ios_lldp_interfaces

It also fixes an issue reported by the same customer related to the vrf_global behavior, where route-target configurations under the address-family context were incorrectly gathered as global VRF attributes.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cisco.ios.vrf_address_family


plugins/module_utils/network/ios/argspec/vrf_address_family/vrf_address_family.py
```diff
@@ -1127,9 +1127,30 @@ class Vrf_address_familyArgs(object):  # pylint: disable=R0903
...
-                                "both": {"type": "str"},
+                                "export": {
+                                    "type": "list",
+                                    "elements": "dict",
+                                    "options": {
+                                        "rt_value": {"type": "str"},
+                                        "stitching": {"type": "bool"},
+                                    },
+                                },
+                                "import_config": {
```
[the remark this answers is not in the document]
KB-perByte: ++

changelogs/fragments/ios_vrf_address_family.yaml
```diff
@@ -0,0 +1,3 @@
+---
+bugfixes:
```
Ruchip16, line 2:
````suggestion
minor_changes:

File: changelogs/fragments/ios_vrf_address_family.yaml [hunk 3]
```diff
@@ -0,0 +1,3 @@
...
+  - ios_vrf_address_family - Enhance facts gathering to correctly parse the `stitching` attribute for route targets and consistently represent `export`/`import_config` as a list of dictionaries.
```
Ruchip16, line 3:
````suggestion
  - ios_vrf_address_family - Added support for parsing the `stitching` attribute under route targets when gathering facts. Enhanced handling of `import_config` and `export` to consistently represent them as lists of dictionaries during fact collection.

**Review by Ruchip16 — CHANGES_REQUESTED** [review]
plz update the changes

File: plugins/module_utils/network/ios/argspec/vrf_address_family/vrf_address_family.py [hunk 4]
```diff
@@ -1127,9 +1127,30 @@ class Vrf_address_familyArgs(object):  # pylint: disable=R0903
                         "route_target": {
                             "type": "dict",
                             "options": {
-                                "export": {"type": "str"},
-                                "import_config": {"type": "str"},
-                                "both": {"type": "str"},
+                                "exports": {
```
KB-perByte, line 1130: ok so we are updating the argspec for 
```
export - exports 
import_config - imports
both - both
```

Now, as it was already released we cannot just go ahead and change the strcutre of them, we should follow a proper deprecation cycle for the attributes going away.

In your case we are almost sorted as it would need minimal changes

Dperecate insted of updating or removing 
```
export
import_config
both
```

Do note that things **that are deprecated in favour of another attribute**, that deprecated attribute should never be a part of facts.
**We should also make sure if someone is using export in their old playbook, config side code should be able to handle it under exports and process it likewise.** ( same for other attributes.

The tricky bit would be the **both** attibute, so deprecate both and name the new attribute to something like **both_options** or AYL.

plugins/module_utils/network/ios/argspec/vrf_global/vrf_global.py
```diff
@@ -64,9 +64,9 @@ class Vrf_globalArgs(object):  # pylint: disable=R0903
                         "route_target": {
                             "type": "dict",
                             "options": {
-                                "export": {"type": "str"},
-                                "import_config": {"type": "str"},
-                                "both": {"type": "str"},
+                                "exports": {"type": "list", "elements": "str"},
```
KB-perByte, line 67: Same as above.

plugins/modules/ios_vrf_address_family.py
```diff
@@ -461,15 +461,42 @@
             description: Specify Target VPN Extended Communities.
             type: dict
             suboptions:
-              export:
+              exports:
```
KB-perByte, line 464: Example on how we do attribute deprecation- https://github.com/ansible-collections/cisco.ios/blob/v10.0.0/plugins/modules/ios_bgp_address_family.py#L230

changelogs/fragments/ios_vrf_address_family.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
```
KB-perByte, line 2: Bugfix please

KB-perByte (changes requested):
Neatly done, please do the attribute deprecation.

Ruchip16 (approved):

PR #1193, opened 2025-05-22
Add default acls

nickbhasin:
##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/network/ios/config/acls/acls.py
+++ b/plugins/module_utils/network/ios/config/acls/acls.py
@@ -43,6 +43,16 @@ class Acls(ResourceModule):
             resource="acls",
             tmplt=AclsTemplate(),
         )
+        self.default_acls = [
+            "implicit_deny_v6",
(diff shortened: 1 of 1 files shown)


plugins/module_utils/network/ios/config/acls/acls.py
```diff
@@ -66,6 +76,10 @@ def generate_commands(self):
         if self.want:
             wantd = self.list_to_dict(self.want)
 
+        for k, have in iteritems(haved):
```
KB-perByte, line 79: let's do the same thing in want too. Rather than iterating it all over again, try to do this within the list-to-dict code.
It is needed in want, so that if the user is putting these in playbook they are not processed.

Also let's be loud and explicit in the docs about this change, and the defaults in consideration.

The doc snippet should be under the acl name attribute in the config.

PR #1196, opened 2025-05-28
Helper Address bugfix for L3 Interface

nickbhasin:
SUMMARY
Added bug fix for Helper Address command for cisco.ios.l3 interface.

ISSUE TYPE
Bugfix Pull Request

COMPONENT NAME
cisco.ios.ios_l3_interfaces

INFO
This change covers the fix for helper address command for l3 interface for ios with examples and tests.


plugins/modules/ios_l3_interfaces.py
```diff
@@ -101,9 +119,6 @@
           unreachables:
             description: Enable sending ICMP Unreachable messages.
             type: bool
-          helper_address:
```
KB-perByte, line 104: We must not remove but deprecate attribute that are not in action

plugins/module_utils/network/ios/config/l3_interfaces/l3_interfaces.py
```diff
@@ -55,13 +55,19 @@ def __init__(self, module):
...
             "ipv4.redirects",
             "ipv4.unreachables",
-            "ipv4.helper_address",
             "ipv4.proxy_arp",
             "ipv6.address",
             "ipv6.autoconfig",
             "ipv6.dhcp",
             "ipv6.enable",
         ]
+        self.complex_parsers = [
```
KB-perByte: I am fine with this, but we can directly call the parsers instead of extending muliple top level variables.

plugins/modules/ios_l3_interfaces.py
```diff
@@ -911,17 +953,17 @@
   returned: when I(state) is C(merged), C(replaced), C(overridden), C(deleted) or C(purged)
   type: list
   sample:
-    - "ip address 192.168.0.3 255.255.255.0"
-    - "ipv6 address dhcp rapid-commit"
-    - "ipv6 address fd5d:12c9:2201:1::1/64 anycast"
+    - sample command 1
```
KB-perByte: why is this reverted?

plugins/modules/ios_l3_interfaces.py
```diff
@@ -911,17 +953,17 @@
...
+    - sample command 3
 rendered:
   description: The provided configuration in the task rendered in device-native format (offline).
   returned: when I(state) is C(rendered)
   type: list
   sample:
-    - "ipv6 address FD5D:12C9:2201:1::1/64"
-    - "ip address 192.168.0.3 255.255.255.0"
-    - "ip address autoconfig"
+    - sample command 1
```
KB-perByte: same as above.

tests/unit/modules/network/ios/test_ios_l3_interfaces.py
```diff
@@ -233,7 +233,7 @@ def test_ios_l3_interfaces_parsed(self):
                       mac-address 0000.0000.0001
                       ip redirects
                       ip mtu 1500
-                      ip helper-address
+                      ip helper-address global 10.0.0.1
```
KB-perByte, line 236: is just `ip helper-address` a valid command? If yes I think we should handle it.

KB-perByte (changes requested):

tests/unit/modules/network/ios/test_ios_l3_interfaces.py
nickbhasin: No it is not a valid command. Example - 
_inserthostname-here(config)#interface gigabitEthernet 3
inserthostname-here(config-if)#ip helper-address
% Incomplete command._

plugins/modules/ios_l3_interfaces.py
nickbhasin: I removed it because the above attribute is invalid. It doesn't work at all.

plugins/modules/ios_l3_interfaces.py
nickbhasin: Not sure how it got changed. reverted them back.

plugins/modules/ios_l3_interfaces.py
nickbhasin: Not sure how it got changed. reverted them back.

plugins/module_utils/network/ios/config/l3_interfaces/l3_interfaces.py
nickbhasin: Helper address is the only list sub option in this interface. For this I had to write the logic of iterating through them, that is why kept them separate to lower the run time complexity. Any future list sub options can be added here as well. If you want I can remove and simply add it in the existing parsers list.

nickbhasin: Updated.

PR #1200, opened 2025-06-04
Update facts/l3_interfaces to handle secondary ipv4 addresses

Tukker-Carson:


Tukker-Carson:
Still learning my way around github. Found an issue where I was unable to use the find/l3_interfaces module on Nexus93180YC_EX switches (running nxos version 7.0(3)I7(6)) due to the script including 'secondary' as part of the netmask and attempting to push it through netmask_to_cdr, part of which parses the netmask to int. This patch removes 'secondary' from the netmask and fixed my issues, though I have not completed sufficient tests for a merge.

Issue #1187, opened 2025-05-09
Bug: cisco.ios.ios_bgp_address_family Issues no neighbor Command in Incorrect Context Causing Configuration Failure


yanivr78:
SUMMARY
The cisco.ios.ios_bgp_address_family module issues the no neighbor command in the wrong context (e.g., under address-family ipv4 instead of the global BGP configuration mode), causing an error when attempting to remove a BGP neighbor.

ISSUE TYPE
- Bug Report

COMPONENT NAME
- cisco.ios.ios_bgp_address_family

ANSIBLE VERSION

ansible [core 2.15.0]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.8/site-packages/ansible
  python version = 3.8.10
  jinja version = 3.1.2
  libyaml = True

COLLECTON VERSION:
cisco.ios 4.2.0

CONFIGURATION:
DEFAULT_HOST_LIST(/etc/ansible/ansible.cfg) = ['/etc/ansible/hosts']
DEFAULT_TIMEOUT(/etc/ansible/ansible.cfg) = 30

OS / ENVIRONMENT
Target Device: Cisco Catalyst 9300
IOS Version: 17.3.3
Control Node OS: Ubuntu 22.04

STEPS TO REPRODUCE
Use the cisco.ios.ios_bgp_address_family module to configure a BGP neighbor.
Remove the neighbor configuration by commenting it out in the playbook.
Run the playbook again.

```
---
- hosts: group1
  gather_facts: no
  connection: ansible.netcommon.network_cli
  vars:
    ansible_network_os: "ios"

  tasks:
  - name: debug - bgp address family 
    debug:
      msg: "testing bgp address family "

  - name: manage bgp address family config
    cisco.ios.ios_bgp_address_family:
      config:
        as_number: 65555
        address_family:
        - afi: ipv4
          networks:
            - address: 192.0.2.3
            - address: 192.0.2.2
          neighbors:
            - neighbor_address: 192.0.2.1
              description: test1
              remote_as: 100
              route_maps:
                - name: test_rm
                  out: true
              # log_neighbor_changes: true
#            - neighbor_address: 192.0.3.1 <<< FIRST CREATE IT THEN UNCOMMENT IT AND RE-RUN
#              description: test2
#              remote_as: 100
#              route_maps:
[...]

PR #1203, opened 2025-06-08
Fix state handling and AF parsing in ios_bgp_address_family

AAYUSH2091:
##### SUMMARY
This update moves the no neighbor command logic out of the address-family block to the correct global BGP context, fixing a CLI syntax error from issue 1187.
It refines the behavior of replaced and overridden states to only manage sub-sections like neighbors when they are explicitly defined in the playbook.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cisco.ios.ios_bgp_address_family

### DIFF
--- a/changelogs/fragments/bgp_address_family.yaml
+++ b/changelogs/fragments/bgp_address_family.yaml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - ios_bgp_address_family - Refined state handling for `replaced` and `overridden` modes and enhanced address-family parsing to accurately differentiate between types such as unicast, multicast, and others.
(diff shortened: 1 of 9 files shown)


KB-perByte (approved):

yanivr78:
Hi,

1) I have installed the patch as so:
 # git fetch origin pull/1203/head:pr-1203
 # git checkout pr-1203
 # pip install -r requirements.txt
 # ansible-galaxy collection build

Then I tested with the following states: overridden/merged/replaced, and still I see the same error:
BGP_ADDRESS_FAM_BUG

For more information on how to reproduce, please see the example here:
https://github.com/ansible-collections/cisco.ios/issues/1187

I

PR #1216, opened 2025-07-01
fix ios_vrf_address_family: Add support for  MDT values attributes operational implementations

AAYUSH2091:
##### SUMMARY
This PR adds support for gathering various MDT (Multicast Distribution Tree) configuration options in the ios_vrf_address_family module.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cisco.ios.ios_vrf_address_family

### DIFF
--- a/plugins/module_utils/network/ios/rm_templates/vrf_address_family.py
+++ b/plugins/module_utils/network/ios/rm_templates/vrf_address_family.py
@@ -1254,7 +1254,7 @@ class Vrf_address_familyTemplate(NetworkTemplate):
                             "mdt": {
                                 "overlay": {
                                     "use_bgp": {
-                                        "set": "{{ true }}",  # Following the send_community pattern
+                                        "set": "{{ true }}",
                                     },
                                 },
                             },


plugins/module_utils/network/ios/rm_templates/vrf_address_family.py
```diff
@@ -1158,6 +1228,76 @@ def __init__(self, lines=None, module=None):
...
+                        '("_" + safi if safi is defined else "_unicast") }}': {
+                            "afi": "{{ afi }}",
+                            "safi": (
+                                "{{ safi if safi is defined else "
+                                "'unicast' }}"
+                            ),
+                            "mdt": {
+                                "overlay": {
+                                    "use_bgp": {
+                                        "set": "{{ true }}",  # Following the send_community pattern
```
nickbhasin: Any specific reason for this comment ?

Ruchip16 (approved):
changes lgtm!

AAYUSH2091:
> but i couldnt see any integration test to validate the change, is there a way to add those if possible?

We can't write integration test as the current CML device doesn't support all these mdt parameters.
Have written unit tests

Ruchip16 (approved):

plugins/module_utils/network/ios/rm_templates/vrf_address_family.py
AAYUSH2091: I have written this for myself will remove it.
as this very good way of handling attributes that are having nested commands supported as well as standalone 
in send_community parser in bgp_global we did it the same way.

PR #1227, opened 2025-08-25
Fix parsing of static routes with interface and distance in gathered state

AAYUSH2091:
##### SUMMARY
Fix ios_static_routes fact parsing when static routes use interface + distance.  
Correctly map interface and distance fields instead of misplacing them.  

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
cisco.ios.ios_static_routes

### DIFF
--- a/changelogs/fragments/cli_test.yml
+++ b/changelogs/fragments/cli_test.yml
@@ -1,0 +1,3 @@
+---
+trivial:
+  - Added result.msg check for command timeout in commit confirmed rollback test.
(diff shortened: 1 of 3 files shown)


tests/integration/labs/multi.yaml
```diff
@@ -1,7 +1,7 @@
 annotations:
-  - border_color: '#808080FF'
-    border_style: ''
-    color: '#FFFFFFFF'
+  - border_color: "#808080FF"
```
KB-perByte: revert and tackle this in a seperate PR please

PR #1232, opened 2025-09-15
Added prompt and answer functionalities  to ios config module

komaldesai13:
##### SUMMARY
This PR enables answering prompt using ios_config module which allows to handle prompt while working with ios device in config mode

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
ios_config

### DIFF
--- a/changelogs/fragments/bgp_address_family.yml
+++ b/changelogs/fragments/bgp_address_family.yml
@@ -1,0 +1,3 @@
+---
+bugfixes:
+  - Fixed an issue where configuration within an address family (ipv6) was ignored by the parser.
--- a/plugins/module_utils/network/ios/rm_templates/bgp_address_family.py
+++ b/plugins/module_utils/network/ios/rm_templates/bgp_address_family.py
@@ -80,7 +80,7 @@ class Bgp_address_familyTemplate(NetworkTemplate):
                 r"""
                 \s\saggregate-address
                 (\s(?P<address>\S+))?
-                (\s(?P<netmask>\S+))?
+                (\s(?P<netmask>\d+\.\d+\.\d+\.\d+))?
                 (\s(?P<as_set>as-set))?
                 (\s(?P<summary_only>summary-only))?
                 (\s(?P<as_confed_set>as-confed-set))?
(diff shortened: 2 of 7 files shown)


plugins/modules/ios_config.py
```diff
@@ -47,7 +47,21 @@
         idempotency and correct diff. Be sure to note the configuration command syntax as
         some commands are automatically modified by the device config parser.
     type: list
-    elements: str
+    elements: raw
```
KB-perByte, line 50: We cannot just update this to something that is not a string equivalent.
We need to either deprecate the parent attribute and then introduce a new attribute to support these options!

KB-perByte, line 50: And we should refrain from using raw, either string or dict that should be taken care of by the above comment,

plugins/modules/ios_config.py
```diff
@@ -316,6 +330,14 @@
       filename: backup.cfg
       dir_path: /home/user
 
+- name: Configure Access Session Attributes while handlening prompt
+  cisco.ios.ios_config:
+    lines:
```
KB-perByte, line 335: can we please call the line as `config_line` as it should be independent from the keywords ios appliances have.

plugins/modules/ios_config.py
```diff
@@ -429,9 +461,14 @@ def save_config(module, result):
...
     backup_spec = dict(filename=dict(), dir_path=dict(type="path"))
+    line_spec = dict(
+        line=dict(type="str", required=False),
+        prompt=dict(type="str", required=False),
+        answer=dict(type="str", required=False),
+    )
     argument_spec = dict(
         src=dict(type="str"),
-        lines=dict(aliases=["commands"], type="list", elements="str"),
+        lines=dict(aliases=["commands"], type="list", elements="raw", options=line_spec),
```
KB-perByte, line 506: not raw, as per the above comment.

docs/cisco.ios.ios_config_module.rst
```diff
@@ -458,6 +458,14 @@ Examples
...
           dir_path: /home/user
 
+    - name: Configure Access Session Attributes while handlening prompt
+      cisco.ios.ios_config:
+        lines:
+          - line: access-session authentication attributes filter-spec include list test_filter
+          - line: access-session accounting attributes filter-spec include list test_filter
+            prompt: "Do you wish to continue[yes]:"
+            answer: "yes"
+        save_when: "changed"
```
KB-perByte, line 517: make sure we add a task outup like we do for resource modules
We are to make those enhancement anyways. Why not just start.

plugins/modules/ios_config.py
KB-perByte: Based on our discussion let's do raw @komaldesai13

plugins/modules/ios_config.py
komaldesai13: addressed it

plugins/modules/ios_config.py
```diff
@@ -47,7 +47,21 @@
         idempotency and correct diff. Be sure to note the configuration command syntax as
         some commands are automatically modified by the device config parser.
     type: list
-    elements: str
+    elements: raw
+    options:
```
KB-perByte:
```suggestion
    suboptions:
```

komaldesai13:
pre-commit.ci autofix

PR #1235, opened 2025-09-24
Fix ios bgp address family parsing for ipv6

AAYUSH2091:
##### SUMMARY
Fix IPv6 address-family support in ios bgp module by updating the parser template, adding relevant unit tests

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ios_bgp_address_family


plugins/module_utils/network/ios/rm_templates/bgp_address_family.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, lines=None, module=None):
                 r"""
                 \s\saggregate-address
                 (\s(?P<address>\S+))?
-                (\s(?P<netmask>\S+))?
+                (\s(?P<netmask>(?!as-set|summary-only|as-confed-set|advertise-map|attribute-map|suppress-map)\d+\.\d+\.\d+\.\d+))?
```
KB-perByte: i am not sure if we need to match pattern here or we can just change the order of the sections to make it just work?

changelogs/fragments/bgp_address_family.yml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
```
KB-perByte: bugfix!

plugins/module_utils/network/ios/rm_templates/bgp_address_family.py
AAYUSH2091: Main Distinctions of IPv6 Addresses
IPv6 does not utilize a distinct netmask such as 255.255.255.0. Rather, it displays the prefix length immediately following the address with a slash, such as /64.
It employs hexadecimal symbols (0-9 and a-f), not limited to digits only.
IPv6 utilizes colons (:) to divide sections instead of dots.
we can reduce a series of zeros by using ::, but only a single time in the address.
Due to these rules, matching IPv6 addresses using regex is distinct and more complicated

AAYUSH2091: let me try with changing the order

Ruchip16 (approved):

PR #1244, opened 2025-10-09
ios_bgp_add_fam - Check idempotent long hash password and fix inf

KB-perByte:
##### SUMMARY

Why is this needed?
```
router bgp 560001
 address-family ipv4 vrf IAMGOINGCRAZY
  neighbor 1.1.1.2 password 7 021E015232511520495706
```
Would get parsed like - 
```
ok: [BATMON] => 
    changed: false
    invocation:
        module_args:
            config: null
            running_config: |-
                router bgp 560001
                 address-family ipv4 vrf IAMGOINGCRAZY
                  neighbor 1.1.1.2 password 7 021E015232511520495706 <<<<<<<<<<<<<<<<<< this
            state: parsed
    parsed:
        address_family:
        -   afi: ipv4
            neighbors:
            -   neighbor_address: 1.1.1.2
                password_options:
                    encryption: 7
                    pass_key: inf <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< to this
            safi: unicast
            vrf: IAMGOINGCRAZY
        as_number: '560001'
```
https://github.com/KB-perByte/netcommon/blob/preserve_templates/plugins/module_utils/network/common/utils.py#L698 << this
```
        if value:
            try:
                return ast.literal_eval(value) <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< this
            except Exception:
                return str(value)
```

would eval it to
```
>>> import ast
>>> ast.literal_eval('021E015232511520495706')
inf
>>>
```

The problem: `ast.literal_eval()` sees `021E015...` and interprets it as scientific notation (the E is an exponent marker)
This becomes `0.21 × 10^15232511520495706`, which is an astronomically large number that Python represents as inf (infinity)

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

ios_bgp_address_family

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/network/ios/config/bgp_address_family/bgp_address_family.py
+++ b/plugins/module_utils/network/ios/config/bgp_address_family/bgp_address_family.py
@@ -386,14 +386,6 @@ class Bgp_address_family(ResourceModule):
                             neib["remote_as"] = str(neib.get("remote_as"))
                         if neib.get("local_as") and neib.get("local_as", {}).get("number"):
                             neib["local_as"]["number"] = str(neib["local_as"]["number"])
-                        if neib.get("password_options", {}).get("pass_key"):
-                            neib["password_options"]["pass_key"] = (
-                                neib.get("password_options", {}).get("pass_key").split("_")[1]
-                                if neib.get("password_options", {})
-                                .get("pass_key")
-                                .startswith("pass_")
-                                else neib.get("password_options", {}).get("pass_key")
-                            )
                         # make dict neighbors dict
                         tmp[neib[p_key[k]]] = neib
                     _af["neighbors"] = tmp
--- a/plugins/module_utils/network/ios/rm_templates/bgp_address_family.py
+++ b/plugins/module_utils/network/ios/rm_templates/bgp_address_family.py
@@ -1614,7 +1614,7 @@ class Bgp_address_familyTemplate(NetworkTemplate):
                                 "neighbor_address": UNIQUE_NEIB_ADD,
                                 "password_options": {
                                     "encryption": "{{ encryption }}",
-                                    "pass_key": "pass_{{ pass_key }}",
+                                    "pass_key": "'{{ pass_key }}'",
                                 },
                             },
                         },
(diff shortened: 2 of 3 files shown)


plugins/module_utils/network/ios/config/bgp_address_family/bgp_address_family.py
```diff
@@ -386,6 +386,14 @@ def _bgp_add_fam_list_to_dict(self, tmp_data):
                             neib["remote_as"] = str(neib.get("remote_as"))
                         if neib.get("local_as") and neib.get("local_as", {}).get("number"):
                             neib["local_as"]["number"] = str(neib["local_as"]["number"])
+                        if neib.get("password_options", {}).get("pass_key"):
```
rohitthakur2590: I believe for  `parsed` state operation, the module returns the RM output directly, which will now show pass_key: "pas_"  instead of the actual encrypted one

plugins/module_utils/network/ios/config/bgp_address_family/bgp_address_family.py
```diff
@@ -386,6 +386,14 @@ def _bgp_add_fam_list_to_dict(self, tmp_data):
...
+                            neib["password_options"]["pass_key"] = (
+                                neib.get("password_options", {}).get("pass_key").split("_")[1]
```
rohitthakur2590: I see we have to recompute `neib.get("password_options", {}).get("pass_key")` multiple times , IMHO we can compute it once and save it in var.

KB-perByte: Removed!!

plugins/module_utils/network/ios/config/bgp_address_family/bgp_address_family.py
KB-perByte: Thanks, got a better idea.

PR #1250, opened 2025-11-07
Ios acls gathered fix

nickbhasin:
##### SUMMARY

Gathered state showing incomplete configuration

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

cisco.ios.ios_acls

##### ADDITIONAL INFORMATION

1. Changed range type from int to str.
2. Fixed the parser to support ipv4, other parameters properly.
3. Added the parser to support ipv6.

```paste below

```

### DIFF
--- a/tests/unit/modules/network/ios/test_ios_acls.py
+++ b/tests/unit/modules/network/ios/test_ios_acls.py
@@ -184,8 +184,8 @@ class TestIosAclsModule(TestIosModule):
                                 ],
                                 name="R1_TRAFFIC",
                             ),
-                        ]
-                    )
+                        ],
+                    ),
                 ],
                 state="merged",
             ),
@@ -783,7 +783,7 @@ class TestIosAclsModule(TestIosModule):
                                 ],
                                 name="R1_TRAFFIC",
                             ),
-                        ]
+                        ],
                     ),
                 ],
                 state="replaced",
@@ -1064,7 +1064,7 @@ class TestIosAclsModule(TestIosModule):
                                 ],
                                 name="R1_TRAFFIC",
                             ),
-                        ]
+                        ],
                     ),
                 ],
                 state="replaced",
@@ -1145,7 +1145,7 @@ class TestIosAclsModule(TestIosModule):
                                 ],
                                 name="R2_TRAFFIC",
                             ),
-                        ]
+                        ],
                     ),
                 ],
                 state="overridden",
@@ -1159,7 +1159,7 @@ class TestIosAclsModule(TestIosModule):
             "ip access-list extended 150",
             "deny tcp 198.51.100.0 0.0.0.255 eq telnet 198.51.110.0 0.0.0.255 eq telnet syn dscp ef ttl eq 10",
             "ipv6 access-list R2_TRAFFIC",
-            "permit ipv6 host 2001:DB8:1234:ABCD::2 2001:DB8:ABCD::/48 sequence 10"
+            "permit ipv6 host 2001:DB8:1234:ABCD::2 2001:DB8:ABCD::/48 sequence 10",
         ]
         self.assertEqual(sorted(result["commands"]), sorted(commands))
 
@@ -1420,7 +1420,7 @@ class TestIosAclsModule(TestIosModule):
                                 ],
                                 name="R1_TRAFFIC",
                             ),
-                        ]
+                        ],
                     ),
                 ],
                 state="rendered",
@@ -1432,7 +1432,7 @@ class TestIosAclsModule(TestIosModule):
             "10 remark remark for acl 110",
             "10 deny tcp 192.0.2.0 0.0.0.255 192.0.3.0 0.0.0.255 eq www syn dscp ef ttl eq 10",
             "ipv6 access-list R1_TRAFFIC",
-            "permit tcp host 2001:DB8:1234:ABCD::2 2001:DB8:ABCD::/48 ack fin sequence 10"
+            "permit tcp host 2001:DB8:1234:ABCD::2 2001:DB8:ABCD::/48 ack fin sequence 10",
         ]
         result = self.execute_module(changed=False)
         self.assertEqual(sorted(result["rendered"]), sorted(commands))
@@ -1829,7 +1829,7 @@ class TestIosAclsModule(TestIosModule):
                                 ],
                                 name="R1_TRAFFIC",
                             ),
-                        ]
+                        ],
                     ),
                 ],
                 state="deleted",
--- a/tests/integration/targets/ios_acls/tests/cli/gathered.yaml
+++ b/tests/integration/targets/ios_acls/tests/cli/gathered.yaml
@@ -44,7 +44,7 @@
                     protocol_options:
                       tcp:
                         ack: true
-                        fin: true 
+                        fin: true
                     source:
                       address: 2001:DB8:AB00::/42
                       port_protocol:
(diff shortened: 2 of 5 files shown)


tests/unit/modules/network/ios/test_ios_acls.py
```diff
@@ -978,13 +1039,13 @@ def test_ios_acls_overridden_idempotent(self):
         self.execute_show_command.return_value = dedent(
             """\
             ip access-list extended 110
-                10 permit tcp 198.51.100.0 0.0.0.255 any eq 22 log (tag = testLog)
+                10 permit tcp 198.51.100.0 0.0.0.255 any eq 22 log
```
KB-perByte: Why is this removed these tests were added for specific usecases.
Please revert it back.

tests/unit/modules/network/ios/test_ios_acls.py
```diff
@@ -1331,7 +1406,7 @@ def test_ios_acls_overridden_remark(self):
             """\
             ip access-list standard test_acl
             ip access-list extended 110
-                10 permit tcp 198.51.100.0 0.0.0.255 any eq 22 log (tag = testLog)
+                10 permit tcp 198.51.100.0 0.0.0.255 any eq 22 log
```
KB-perByte: same as above

tests/unit/modules/network/ios/test_ios_acls.py
```diff
@@ -2376,15 +2451,15 @@ def test_ios_parsed_complex_2(self):
...
+                    20 permit object-group BT-ports object-group AIT-BT-Servers
                     30 permit tcp any object-group CP-Appliances eq 443
-                    40 deny object-group deny-mgmt-ports any any
-                    50 permit object-group icmp any object-group IB-Servers
-                    60 permit object-group generic-any-port any object-group IB-Servers
-                    70 permit object-group generic-any-port any object-group AD-Servers
-                    80 permit object-group generic-any-port any object-group CP-Appliances
-                    90 permit object-group generic-any-port any object-group Tanium-Servers
-                    100 permit object-group generic-any-port any object-group SOC-Tenable-Servers
+                    40 deny object-group deny-mgmt-ports any
```
KB-perByte, line 2735: why are we removing any from all the instance is it a wrong command for the current applaince we support?

plugins/module_utils/network/ios/argspec/acls/acls.py
```diff
@@ -55,8 +55,8 @@ class AclsArgs(object):  # pylint: disable=R0903
                                                 "neq": {"type": "str"},
                                                 "range": {
                                                     "options": {
-                                                        "end": {"type": "int"},
-                                                        "start": {"type": "int"},
+                                                        "end": {"type": "str"},
```
KB-perByte, line 58: If the range attribute worked somewhere for some command, and we are updating the type then its a customer facing breaking change.
As in YAML
```
range:
  end: 10
  start: 3
```

Would not be same as 
```
range:
  end: "10"
  start: "3"
```

plugins/module_utils/network/ios/facts/acls/acls.py
```diff
@@ -132,20 +132,18 @@ def populate_facts(self, connection, ansible_facts, data=None):
...
-                        ipv6_add = ace[typ].pop("ipv6_address")
-                        ace[typ]["address"] = ipv6_add.split(" ")[0]
-                        ace[typ]["wildcard_bits"] = ipv6_add.split(" ")[1]
-                    else:
-                        ace[typ]["address"] = ace[typ].pop("ipv6_address")
+                    ace[typ]["address"] = temp.get("address")
+                    ace[typ]["wildcard_bits"] = temp.get("wildcard_bits")
 
             def process_protocol_options(each):
+                ICMP_MAP = {
```
KB-perByte: We are handling only one?
https://github.com/ansible-collections/cisco.ios/blob/main/plugins/module_utils/network/ios/config/acls/acls.py#L392 Can this be used instead?

KB-perByte (changes requested):

tests/unit/modules/network/ios/test_ios_acls.py
KB-perByte: For your reference, https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/sec_data_acl/configuration/12-4t/sec-data-acl-12-4t-book/sec-acl-syslog.html
image
Over a search I got it for cisco ios v12.x but it should be there for the versions we support

plugins/module_utils/network/ios/argspec/acls/acls.py
nickbhasin: If we enter end as 23 then the command present in the device has telnet instead of the number. To support that I have changed it to str.

plugins/module_utils/network/ios/facts/acls/acls.py
nickbhasin: This task is in TODO. This PR only fixes the customer's issue for now. It will be picked later in the sprint. I have just added a custom MAP since the original map is not doing conversion properly.

tests/unit/modules/network/ios/test_ios_acls.py
nickbhasin: None of our devices supported it. Will fix it. - 
Screenshot From 2025-11-07 14-22-43
Screenshot From 2025-11-07 14-23-31
Screenshot From 2025-11-07 14-26-43

tests/unit/modules/network/ios/test_ios_acls.py
nickbhasin: Yes. It can either be any or object group or (ipv4 address + ipv4 wildcard bits)

tests/unit/modules/network/ios/test_ios_acls.py
KB-perByte: so this option wont be present as is. But tag is visible in this pattern when we read the configuration back from the device.
We might have to check how to add tags. Please refer cisco docs around it.

tests/unit/modules/network/ios/test_ios_acls.py
KB-perByte: ack

plugins/module_utils/network/ios/facts/acls/acls.py
KB-perByte: If the fix for this in TODO, then no traces of this should exist in this PR which targets fix for the customer issue.
And if the customer fix needs a fix in the custom map then we fix it the right way in this PR iteslf.

plugins/module_utils/network/ios/argspec/acls/acls.py
KB-perByte: So we fix things for telnet what about any other range that was already supported by the module?

tests/unit/modules/network/ios/test_ios_acls.py
nickbhasin: Reverted the changes as discussed.

tests/unit/modules/network/ios/test_ios_acls.py
nickbhasin: Reverted the changes as discussed.

plugins/module_utils/network/ios/argspec/acls/acls.py
nickbhasin: Keeping it the same as it is not impacting any existing playbook and is necessary to support range names like telnet.

plugins/module_utils/network/ios/facts/acls/acls.py
nickbhasin: Removed the TODO as discussed.

AAYUSH2091 (approved):

PR #1257, opened 2025-11-17
Sonar Integration

nickbhasin:
##### SUMMARY

Sonar Integration

### DIFF
--- a/.github/workflows/sonarcloud.yaml
+++ b/.github/workflows/sonarcloud.yaml
@@ -1,5 +1,6 @@
+---
 name: SonarCloud
-on: # yamllint disable-line rule:truthy
+on:  # yamllint disable-line rule:truthy
   push:
     branches: [main]
   pull_request_target:


.github/workflows/sonarcloud.yaml
```diff
@@ -0,0 +1,134 @@
...
+on: # yamllint disable-line rule:truthy
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  codecoverage:
+    name: "Code Coverage | Python 3.12"
+    runs-on: ubuntu-latest
```
KB-perByte, line 12: https://redhat-internal.slack.com/archives/C9W5WCSMS/p1763381845466859?thread_ts=1763111503.059599&cid=C9W5WCSMS 

We discussed that we don't want to go with 2 runners, and this change only simplifies the changes to one file but still keeps 2 jobs
Is it necessary?

.github/workflows/sonarcloud.yaml
```diff
@@ -0,0 +1,134 @@
+---
+# With much help from:
```
KB-perByte: Why is this needed??

.github/workflows/sonarcloud.yaml
```diff
@@ -0,0 +1,134 @@
...
+
+      - name: Set up Python ${{ env.python_version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.python_version }}
+
+      - name: Install Python + testing dependencies
+        run: |
+          python3 -m pip install --upgrade pip
+          python3 -m pip install ansible-core pytest pytest-cov pytest-ansible-units pytest-forked pytest-xdist
```
KB-perByte, line 33: Do we always test with the latest release?

KB-perByte: of ansible-core?

KB-perByte (changes requested):

.github/workflows/sonarcloud.yaml
nickbhasin: Following similar pattern - 
https://github.com/ansible/aap-dev/blob/main/.github/workflows/sonarcloud.yaml

If you want I can remove it.

.github/workflows/sonarcloud.yaml
nickbhasin: As per the JIRA - https://issues.redhat.com/browse/ACA-4431
We should actually have 2 separate files with 2 separate runners as it is asking for the following - 

- CI runs and uploads the coverage artifact.
- SonarCloud workflow triggers post-CI.
- SonarCloud PR decoration model should be followed.

.github/workflows/sonarcloud.yaml
nickbhasin: Yes. following the current flow in our main branch calls this reusable workflow - 
https://github.com/ansible-network/github_actions/blob/main/.github/workflows/coverage_network_devices.yml

.github/workflows/sonarcloud.yaml
```diff
@@ -0,0 +1,78 @@
...
+        run: |
+          gh pr checkout ${{ github.event.pull_request.number }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: SonarCloud Scan
+        uses: SonarSource/sonarqube-scan-action@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
```
gundalow: I think this is what you need

```suggestion
          SONAR_TOKEN: ${{ secrets.ANSIBLE_COLLECTIONS_ORG_SONAR_TOKEN_CICD_BOT }}
```

.github/workflows/sonarcloud.yaml
nickbhasin: Keeping it 1 file with 1 runner as discussed.

.github/workflows/sonarcloud.yaml
nickbhasin: Updated.

.github/workflows/sonarcloud.yaml
nickbhasin: removed.

PR #1326, opened 2026-04-15
ios_acl : handling of  protocol type values in ACL configuration

komaldesai13:
## Description

- What is being changed?

  1. fix in rm_templates/acls.py (lines 422, 561) —
   Changed the template so icmp_igmp_protocol is only used
  when protocol is icmp or igmp:                              
  {{ tcp_flags if tcp_flags is defined else                 
  (icmp_igmp_protocol if protocol is defined and protocol in  
  ['icmp', 'igmp'] else undefined) }}

  2. Defensive check in facts/acls/acls.py — Changed if/if/if 
  to if/elif chain, and added isinstance(protocol_options,    
  str) check before calling .split() on TCP protocol options
  as a safety net.            
             
- Why is this change needed?
  ios_acls module crashed with 'int' object has no   
  attribute 'split' on some hosts during state: gathered /
  state: overridden. The regex group icmp_igmp_protocol (which
   includes \d+) was capturing trailing numeric values for all
   protocols, including TCP. When tcp_flags wasn't defined,   
  those numbers leaked into protocol_options and .split() was 
  called on an int.    

                                                              

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Collection release
- [ ] CI maintenance
- [ ] Workflow maintenance
- [ ] Configuration change

## Related Issue

- Fixes #https://redhat.atlassian.net/browse/AAP-69108
- Related to #

## Component Name
ios_acl

## Self-Review Checklist

- [ ] I have performed a self-review of my code
- [ ] I have removed all commented code (no commented code should be merged)
- [ ] I have updated documentation where needed
- [ ] I have considered the security impact of these changes
- [ ] I have considered performance implications
- [ ] I have thought about error handling and edge cases
- [ ] I have tested the changes in my local environment
- [ ] I have verified the changes work with the target IOS version(s)
- [ ] I have reviewed the acceptance criteria for related tickets

### Prerequisites

- IOS Version:
- Hardware Platform (if applicable):
- Test Topology:

### Steps to Test
1.
2.
3.

### Test Results

```
```

- [ ] All acceptance criteria have been reviewed and verified
- [ ] Acceptance criteria checklist:
  - [ ]
  - [ ]
  - [ ]

### Command Output / Logs

```
<!-- Paste below -->
```

### Required Actions

- [ ] Requires documentation updates
  
- [ ] Requires changelog fragment
  
- [ ] Requires integration test updates
- [ ] Requires unit test updates
- [ ] Requires coordination with other teams
- [ ] Blocked by PR/MR: #XXX

### DIFF
--- a/plugins/module_utils/network/ios/facts/acls/acls.py
+++ b/plugins/module_utils/network/ios/facts/acls/acls.py
@@ -172,18 +172,23 @@ class AclsFacts(object):
                     protocol = each_ace.get("protocol")
                     protocol_options = each_ace.get("protocol_options")
                     if protocol_options is not None:
-                        if protocol == "tcp":
+                        # Skip if already processed into dict format
+                        if isinstance(protocol_options, dict):
+                            # Already processed (protocol_number case) or structured data
+                            pass
+                        elif protocol == "tcp":
                             # Handle TCP protocol options - should be string flags like 'ack fin'
-                            # If numeric, it's invalid for TCP, so remove it
-                            if isinstance(protocol_options, int):
-                                each_ace.pop("protocol_options", None)
-                            else:
+                            # If numeric or non-string, it's invalid for TCP, so remove it
+                            if isinstance(protocol_options, str):
                                 flag_dict = {flag: True for flag in protocol_options.split()}
                                 each_ace["protocol_options"] = {"tcp": flag_dict}
-                        if protocol == "icmp":
+                            else:
+                                # Invalid type (int or other) for TCP flags
+                                each_ace.pop("protocol_options", None)
+                        elif protocol == "icmp":
                             protocol_options = str(protocol_options).replace("-", "_")
                             each_ace["protocol_options"] = {"icmp": {protocol_options: True}}
-                        if protocol == "igmp":
+                        elif protocol == "igmp":
                             protocol_options = str(protocol_options).replace("-", "_")
                             mapping = IGMP_MAP.get(protocol_options, protocol_options)
                             each_ace["protocol_options"] = {"igmp": {mapping: True}}
--- a/plugins/module_utils/network/ios/facts/acls/acls.py
+++ b/plugins/module_utils/network/ios/facts/acls/acls.py
@@ -172,18 +172,11 @@ class AclsFacts(object):
                     protocol = each_ace.get("protocol")
                     protocol_options = each_ace.get("protocol_options")
                     if protocol_options is not None:
-                        # Skip if already processed into dict format
-                        if isinstance(protocol_options, dict):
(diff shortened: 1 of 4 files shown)


plugins/module_utils/network/ios/facts/acls/acls.py
```diff
@@ -172,13 +172,23 @@ def process_protocol_options(each):
...
                     if protocol_options is not None:
-                        if protocol == "tcp":
-                            flag_dict = {flag: True for flag in protocol_options.split()}
-                            each_ace["protocol_options"] = {"tcp": flag_dict}
-                        if protocol == "icmp":
+                        # Skip if already processed into dict format
+                        if isinstance(protocol_options, dict):
+                            # Already processed (protocol_number case) or structured data
+                            pass
+                        elif protocol == "tcp":
```
nickbhasin, line 179: How are we getting protocol_options as dict ? Also, we dont allow numerical values for tcp as per argspec then how are we getting an integer. It seems the issue is with the parser.

komaldesai13: updated code to address issue at parser level first and then added double check in facts file also

plugins/module_utils/network/ios/rm_templates/acls.py
```diff
@@ -558,7 +560,9 @@ def __init__(self, lines=None):
                                 "evaluate": "{{ evaluate }}",
                                 "protocol": "{{ protocol }}",
                                 "protocol_number": "{{ protocol_num }}",
-                                "protocol_options": "{{ tcp_flags if tcp_flags is defined else icmp_igmp_protocol }}",
+                                "protocol_options": "{{ tcp_flags if tcp_flags is defined"
+                                " else (icmp_igmp_protocol if protocol is defined"
+                                " and protocol in ['icmp', 'igmp'] else undefined) }}",
```
TheNova22, line 565: Is there a reason we are marking it as undefined and not None like rest of the templates defined in this file?

Example: https://github.com/ansible-collections/cisco.ios/blob/main/plugins/module_utils/network/ios/rm_templates/acls.py#L478

komaldesai13: undefined is used intentionally here — it behaves           
  differently from None in this context          
                                            
  - `else None `renders as the string "None", so the            
  protocol_options key would remain in the parsed output with 
  a truthy value, causing incorrect protocol options to appear
   in the config.                                             
  - `else undefined ' renders as an empty string, which causes
  NetworkTemplate to omit the key entirely from the result.

TheNova22 (approved):

Issue #1309, opened 2026-03-18
Idempotence not working for SSH keys in module 'ios_user'


Peve753:
### Bug Summary

This [function](https://github.com/ansible-collections/cisco.ios/blob/c3f336885eb7c31cd77a9b0648fd0bbc0c372822/plugins/modules/ios_user.py#L690-L715) receives the right details from cisco switches in the "data" variable but doesn't properly format them for further necessary comparison leaving variables (like "sshkey") empty.

Meaning idempotency for SSH keys is not guaranteed and the task runs without actual changes.

### Select the relevant components

- [x] Module
- [ ] Action Plugin
- [ ] Connection Plugin (cliconf)
- [ ] Terminal Plugin
- [ ] Module Utils
- [ ] Documentation
- [ ] Integration Tests
- [ ] Unit Tests
- [ ] Collection Release
- [ ] CI Maintenance
- [ ] Workflow Maintenance

### IOS Version

IOS-XE 17.18.02 and others

### Hardware Platform

Catalyst 9200L-48P-4X and others

### Module/Plugin Name

ios_user

Peve753:
Edit:
This line [Regex_findall](https://github.com/ansible-collections/cisco.ios/blob/64a7bd9fd05062e1b3d9c7d302881a482a9415ca/plugins/modules/ios_user.py#L692) only filters patterns for usernames that start immediately at line start or after exactly 2 whitespace chars. It misses usernames that start after 1 whitespace char.
Example:
```
\nusername ✓
\n  username ✓
\n username ❌ <-- will be ignored
```

Just including the 1 whitespace case fixes this
`re.findall("(?:^(?:u|\\s{1,2}u))sername (\\S+)", data, re.M)`

Found with IOS XE Version 17.15.04

PR #1328, opened 2026-04-20
ios_user | Feature addition to handle multiple sshkeys

### LINKED ISSUE #1322: Adding SSH Pubkey Want/ Have check for `ios_user` module
### Feature Summary

As cisco set limitations of a max. of 2 ssh keys per user on their switches old keys have to be explicitly removed and will not be automatically deleted by e.g. adding a third key.

I want to suggest to implement a small comparison between "have"-state and "want"-state for ssh keys following deletion of any leftover keys. 

Without it if user "foo" already holds 2 keys on a switch (key "A" and key "B") and the playbook knows key "A" and a third (key "C") and tries to set them key "C" will be rejected as user "foo" exceeded his limit.

This scenario already applies for old/ unused users as the module deletes them completely alongside their keys. But if it's a known and active user and they just keep changing their key the old ones are stuck and preventing new ones to be set on the switch unless deleted manually.

The source code already translates the input key to MD5 fingerprints for comparison which is how the switch outputs them via "show running-conf".

That way both sides have the exact same state at all times.

Image

- [ ] New Module
- [x] Enhancement to Existing Module
- [ ] New Action Plugin
- [ ] Enhancement to Existing Action Plugin
- [ ] Terminal Plugin
- [ ] Module Utils
- [ ] Documentation
- [ ] Collection Release
- [ ] CI Maintenance
- [ ] Workflow Maintenance

TheNova22:
## Description

- What is being changed?

The `ios_user` module is being enhanced with a new **purge_keys** parameter that allows automatic removal of SSH keys not specified in the `sshkey` parameter. 
Key changes include:
Added purge_keys parameter (boolean, default: false) to ios_user.py
New remove_ssh() function that removes SSH keys not in the wanted configuration
Validation to enforce Cisco IOS's 2-key limit per user
Updated documentation and examples showing the new functionality
Comprehensive unit and integration tests for the purge_keys feature
Minor regex improvements for parsing SSH key configurations

- Why is this change needed?

Cisco IOS devices have a maximum limit of 2 SSH keys per user. Without this feature:
Users cannot easily replace existing SSH keys when the limit is reached
Manual removal of old keys is required before adding new ones
Key rotation workflows become cumbersome and error-prone
Idempotency is difficult to achieve when managing SSH keys

- How does this change address the issue?

The `purge_keys` parameter solves this by:
**Automatic key cleanup**: When `purge_keys: true` and `state: present`, it removes keys present on the device that aren't in the sshkey list
**Enables key rotation**: Users can specify new keys and have old ones automatically removed
**Respects the 2-key limit**: Added validation that fails if more than 2 keys are supplied
**Safety measures**: Warns if attempting to purge admin user keys; only works with state: present
**Maintains idempotency**: Running the same playbook multiple times produces no changes after initial application
Example workflow from the tests:

```
- name: Replace SSH keys
  cisco.ios.ios_user:
    name: ansible
    sshkey:
      - "{{ lookup('file', '~/new_key.pub') }}"
    purge_keys: true  # Removes old keys, adds new one
```

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Test update
- [ ] Collection release
- [ ] CI maintenance
- [ ] Workflow maintenance
- [ ] Configuration change

## Related Issue

- Fixes #1322 

## Component Name

ios_user

## Self-Review Checklist


TheNova22:
Tests are successful - Here are the items covered:
Unit
- Behavior when purge keys is set to true and false
- Behavior when purge keys is set to true but keys already exist (idempotency)
- Behavior when more than 3 keys are passed (failure)

Integration
- 0 keys to 2 Keys Addtion (No Removal)
- Idempotency
- 2 Keys to new 1 Key (removes 2 keys)
- Add another key (removes none, adds one)
- Failure with three keys
- Aggregate behavior

nickbhasin (changes requested):
Integration TCs, sonar analysis is failing. Please fix and re raise for review.

alehaa:
I believe this will also resolve issue #1309, as it now checks multiple levels of indentation for the `username`.

TheNova22:
The mentioned sonarcloud comments should be handled as a separate enhancement as they require updation of the main functions used by the module. The current PR focuses on using the existing methods and adding another feature to it.

plugins/modules/ios_user.py
```diff
@@ -582,6 +646,25 @@ def user_del_cmd(username):
...
+def remove_ssh(command, want, have):
+    """
+    Function responsible to add no key-hash commands
+    Compares want and have via set operations
+    """
+    want_keys_set = set(want.get("sshkey", []))
+    have_keys_set = set(have.get("sshkey", []))
+
+    keys_to_be_removed = have_keys_set.difference(want_keys_set)
+    # Nothing to be removed if wanted keys are same as keys present
```
nickbhasin: Remove this comment.

plugins/modules/ios_user.py
nickbhasin, line 824: Resolve this sonar cloud warning on the sonar portal as of now if not possible through code.

nickbhasin (approved):
Approving. Merge it only when the CI is green.

changelogs/fragments/ios_user_sshkey_purging-major_changes.yaml
```diff
@@ -0,0 +1,5 @@
+---
+major_changes:
+  - ios_user module adds purge_keys parameter to manage multiple SSH keys per user.
+    Cisco IOS devices support maximum 2 SSH keys per user. The purge_keys parameter
```
rohitthakur2590: why this is major change ? (as breaking changes are the one which comes under major_changes category)

TheNova22: Oh, my bad, i had assumed a new feature addition would fall under major_change
Will bump it down to minor_changes

komaldesai13 (approved):

PR #1340, opened 2026-05-12
Fix #1339: Allow whitespace on config line ending

### LINKED ISSUE #1339: New 'purge_keys' parameter in 'ios_user' module cannot compare ssh-keys correctly
### Bug Summary

In testing the newly included 'purge_keys' parameter i noticed a bug that currently prevents all our machines from using this feature as it is unable to detect the keys currently stored on the switch. Both on IOS and IOS XE.

After debugging locally i found the issue. Within the "parse_sshkey" function a regex filter is having problems finding matches when comparing it to the playbook ssh-keys.

In the picture you can see an example with the same format that our cisco machines would provide using this module.
Once it tries to search for matches most of them return empty, except 1.

Image

That one exception is missing a whitespace at the end between which is the reason the filter is having trouble here. All other lines include a whitespace after ssh-key and "\n" that's not being covered in the regex filter.

After removing that one whitespace the function behaves as intended.

Image

### Solution

Covering a potential whitespace here would fix the issue like this: `key-hash\s+(\S+\s+\S+(?:\s+\S+)?)\s*$`

- [x] Module
- [ ] Action Plugin
- [ ] Terminal Plugin
- [ ] Module Utils
- [ ] Documentation
- [ ] Integration Tests
- [ ] Unit Tests
- [ ] Collection Release
- [ ] CI Maintenance
- [ ] Workflow Maintenance

### IOS Version

_No response_

### Hardware Platform

_No response_

### Module/Plugin Name

ios_user

alehaa:
## Description

IOS and IOS-XE might have a whitespace on `key-hash` lines before line ending. This MR fixes the regex to allow an optional whitespace if needed.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Collection release
- [ ] CI maintenance
- [ ] Workflow maintenance
- [ ] Configuration change

## Related Issue

- Fixes #1339 
- Related to #1328 

## Component Name

ios_user

## Self-Review Checklist

- [ ] I have updated documentation where needed
- [ ] I have considered the security impact of these changes
- [ ] I have considered performance implications
- [ ] I have reviewed the acceptance criteria for related tickets

### Prerequisites

- IOS Version:
- Hardware Platform (if applicable):
- Test Topology:

### Steps to Test
1.
2.
3.

### Test Results

```
```

- [ ] All acceptance criteria have been reviewed and verified
- [ ] Acceptance criteria checklist:
  - [ ]
  - [ ]
  - [ ]

### Command Output / Logs

```
<!-- Paste below -->
```

### Required Actions

- [ ] Requires documentation updates
  
- [ ] Requires changelog fragment
  
- [ ] Requires integration test updates
- [ ] Requires unit test updates
- [ ] Requires coordination with other teams
- [ ] Blocked by PR/MR: #XXX


TheNova22:
Thanks for the quick contribution @alehaa!

Could you please add a changelog(you can refer to any non-release PR) and add or update a testcase that checks that whitespace scenarios are working.

alehaa:
@TheNova22 I've added a fragment for the changelog. However, I'm not sure how to proceed with the test case. I would like to add a second SSH key to the testing configuration file in tests/unit/modules/network/ios/fixtures/ios_user_config.cfg. Since pre-commit.ci is active, I assume the trailing whitespace required for the test case would be removed again.

TheNova22:
That's a good point @alehaa! Please check the following out and you can verify if this work for you:

Update .pre-commit-config.yaml to include the path of interest with key as exclude under [trailing-whitespace](https://github.com/ansible-collections/cisco.ios/blob/main/.pre-commit-config.yaml#L18).
It is a single string entity that takes in regex input. So it will look like the following:
`        exclude: 'tests/unit/modules/network/ios/fixtures/ios_user_config\.cfg$'`

https://pre-commit.com/#config-exclude

I validated on my local machine and it seems to work.

alehaa:
@TheNova22 I've updated the test cases as intended. Although I couldn't find any IOS or IOS‑XE device in our deployment that has key hashes without trailing whitespace, I didn't remove the existing test case so it covers both kinds of configurations, e.g. if a local configuration file removes the trailing whitespace.

However, I'm encountering an error in the `test_ios_user_set_sshkey_purge_keys_true` test case: because the `ios_user` module implementation uses sets, the order of key hashes is unstable. Should I change this implementation to use a stable list, or should I match the commands using `sorted()`, i.e. just verify that every command is issued but out of order?

TheNova22:
Hi @alehaa 
Thanks again for checking it out and including both the scenarios(with and without trailing whitespace).

Going with the `verify that every command is issued but out of order` seems fine with me as the hashes of each key will be different, so if they appear, then we should be good.
And yes, you can match them by either using `sorted`, which is used across different testcases like [here](https://github.com/ansible-collections/cisco.ios/blob/main/tests/unit/modules/network/ios/test_ios_vrf_address_family.py#L584) or you can use `assertCountEqual` as well. Both should achieve the same result according to me.

alehaa:
Alright, then this PR should be ready to go. Feel free to squash everything together during the merge. ;)

TheNova22 (approved):

.pre-commit-config.yaml
```diff
@@ -16,6 +16,7 @@ repos:
       - id: no-commit-to-branch
         args: [--branch, main]
       - id: trailing-whitespace
```
rohitthakur2590, line 18: anyone looking at this later will wonder why that one file is excluded. A comment  something like below would help.
# ios_user_config.cfg intentionally contains a trailing space on a key-hash
  # line to simulate IOS device output. See PR #1340.

rohitthakur2590 (changes requested):

komaldesai13 (approved):

rohitthakur2590 (approved):

PR #1341, opened 2026-05-18
fix(ios_acls): generate correct IPv6 ACL remark commands

### LINKED ISSUE #1336: IPv6 ACL remarks missing "sequence" keyword in ios_acls module
## Summary
When using `cisco.ios.ios_acls` with IPv6 ACLs and remarks, 
the module generates wrong CLI commands.

## Ansible Version
ansible core 2.19.4

## Collection Version
cisco.ios 11.3

## Steps to Reproduce
Using ios_acls with state: replaced and an IPv6 ACL containing remarks:

    - name: Deploy ACL
      cisco.ios.ios_acls:
        config:
          - afi: ipv6
            acls:
              - name: TEST_ACL
                aces:
                  - sequence: 10
                    remarks:
                      - Test remark
        state: replaced

## Expected Behavior
Generated command should be:
    sequence 10 remark Test remark

And for negation:
    no sequence 10 remark

## Actual Behavior
Generated command is:
    10 remark Test remark

And for negation:
    no 10 remark

This causes IOS to return:
    % Invalid input detected at '^' marker.

## Root Cause
Two bugs found in the source:

**rm_templates/acls.py:**
1. `remarks_with_sequence` function does not differentiate between 
   IPv4 and IPv6 - missing "sequence" keyword for IPv6
2. `remarks_ipv6` parser has wrong setval - missing sequence in command
3. Missing function for "no remarks" handling in IPv6

**config/acls/acls.py:**
4. `afi` not passed to `addcmd` calls for remarks

### rm_templates/acls.py

```python
def remarks_with_sequence(remarks_data):
    cmd = "remark"
    if remarks_data.get("remarks"):
        cmd += " " + remarks_data.get("remarks")
    if remarks_data.get("sequence"):
        if remarks_data.get("afi") == "ipv6":
            cmd = "sequence " + to_text(remarks_data.get("sequence")) + " " + cmd
        else:
            cmd = to_text(remarks_data.get("sequence")) + " " + cmd
    return cmd

def remarks_no_data_with_sequence(remarks_data):
    if remarks_data.get("afi") == "ipv6":
        return "sequence " + to_text(remarks_data.get("sequence")) + " remark"
    else:
        return to_text(remarks_data.get("sequence")) + " remark"
```
### Change remarks_ipv6 parser setval:
`"setval": remarks_with_sequence`

### Change remarks_no_data parser setval:
` "setval": remarks_no_data_with_sequence `

### config/acls/acls.py
Add afi to all addcmd calls for remarks:
``` 
self.addcmd(
    {
        "sequence": hentry.get("sequence", None),
        "afi": afi,
    },
    "remarks_no_data",
    negate=True,
)
``` 

- [x] Module
- [ ] Action Plugin
- [ ] Terminal Plugin
- [ ] Module Utils
- [ ] Documentation
- [ ] Integration Tests
- [ ] Unit Tests
- [ ] Collection Release
- [ ] CI Maintenance
- [ ] Workflow Maintenance

### IOS Version

17.12.4

### Hardware Platform

Catalyst

### Module/Plugin Name

ios_acls

malikdotexe:
## Summary

Fixes incorrect CLI command generation for IPv6 ACL remarks in the `ios_acls` module.

**Before (broken):**
- Add remark: `10 remark Test remark`
- Remove remark: `no 10 remark`

**After (correct):**
- Add remark: `sequence 10 remark Test remark`
- Remove remark: `no sequence 10 remark`

The IPv4-style bare-number format causes IOS to return `% Invalid input detected at '^' marker` when applied to IPv6 ACLs.

Closes #1336

---

## Root Cause Analysis

Four interconnected bugs were identified:

### 1. `remarks_with_sequence()` was not AFI-aware (`rm_templates/acls.py`)

The function always generated `{sequence} remark {text}` regardless of address family. IPv6 ACLs on IOS require the literal keyword `sequence` before the number (`sequence {N} remark {text}`), while IPv4 uses the bare number (`{N} remark {text}`).

### 2. `remarks_ipv6` template had a broken `setval` (`rm_templates/acls.py`)

The `getval` regex **correctly** parsed IPv6 remarks (`sequence \d+ remark .+`) — proving the developers knew the correct format — but the `setval` was a plain Jinja2 string `"remark {{ remarks }}"` that omitted the sequence number entirely. This template could never generate a valid command.

### 3. No IPv6-aware negation handling

When negating remarks, the module always used the `remarks_no_data` template which calls `remarks_with_sequence()`. Since that function had no AFI awareness (bug #1), negation always produced the IPv4 format `no {N} remark` instead of `no sequence {N} remark`.

### 4. `afi` not passed to remark `addcmd` calls (`config/acls/acls.py`)

In `_compare_aces()`, the `add_afi()` helper is used for ACE entries but all four `addcmd` calls for remarks constructed their data dicts with only `sequence` (and optionally `remarks`) — never including `afi`. Even if the template function were AFI-aware, it would never receive the information needed to branch.

---

## Fix Approach

Rather than introducing a second function (as suggested in the issue), this PR uses a **single-function approach** that is simpler and avoids code duplication:

1. **`remarks_with_sequence()`** — Added an `afi` check: when `afi == "ipv6"`, prepend `"sequence "` before the number. The function naturally handles both the "with remark text" and "no data" (negation) cases since when `remarks` is absent it simply produces `sequence {N} remark` (or `{N} remark` for IPv4). No second function needed.

2. **`remarks_ipv6` template** — Changed `setval` from the broken Jinja2 string to `remarks_with_sequence` (the now-AFI-aware function).

3. **`_compare_aces()` in `config/acls/acls.py`** — Added `"afi": afi` to all four remark-related `addcmd` data dicts so the template function receives the address family.

### Why not a separate `remarks_no_data_with_sequence` function?

The issue suggested creating a dedicated function for the negation case. This is unnecessary because `remarks_with_sequence()` already handles it — when the `remarks` key is absent/falsy, the function outputs `sequence {N} remark` (IPv6) or `{N} remark` (IPv4) with no trailing text. The framework then prepends `no` for negation. One function, zero duplication, same correctness.

---

## Key Insight

The `remarks_ipv6` template's `getval` regex already correctly matches `sequence \d+ remark .+`. This is **proof** that IOS uses this syntax and that the parsing side was tested (parsing fires every time the module reads device state). The `setval` side was simply never implemented correctly — it only fires when making changes, and IPv6 remark modifications were never tested against a real device.

---

## Testing

- Updated existing unit test (`test_ios_acls_merged_remarks_positional`) to expect correct `sequence N remark` format for IPv6 ACLs
- Added new unit test (`test_ios_acls_replaced_ipv6_remarks`) that verifies both remark addition and negation use the `sequence` keyword for IPv6
- All 36 unit tests pass
- Sanity tests pass

## Changelog

Added `changelogs/fragments/fix_ipv6_acl_remarks.yaml` with bugfix entry.

### DIFF
--- a/changelogs/fragments/fix_ipv6_acl_remarks.yaml
+++ b/changelogs/fragments/fix_ipv6_acl_remarks.yaml
@@ -1,0 +1,6 @@
+---
+bugfixes:
+  - ios_acls - Fix incorrect CLI command generation for IPv6 ACL remarks.
+    The module now correctly generates ``sequence N remark`` syntax for IPv6
+    instead of the IPv4-style ``N remark`` format. Negation also correctly
+    uses ``no sequence N remark``.
--- a/plugins/module_utils/network/ios/config/acls/acls.py
+++ b/plugins/module_utils/network/ios/config/acls/acls.py
@@ -210,6 +210,7 @@ class Acls(ResourceModule):
                         self.addcmd(
                             {
                                 "sequence": hentry.get("sequence", None),
+                                "afi": afi,
                             },
                             "remarks_no_data",
                             negate=True,
@@ -236,6 +237,7 @@ class Acls(ResourceModule):
                         self.addcmd(
                             {
                                 "sequence": hentry.get("sequence", None),
+                                "afi": afi,
                             },
                             "remarks_no_data",
                             negate=True,
@@ -245,6 +247,7 @@ class Acls(ResourceModule):
                             {
                                 "remarks": wrems,
                                 "sequence": wentry.get("sequence", ""),
+                                "afi": afi,
                             },
                             "remarks",
                         )
@@ -264,6 +267,7 @@ class Acls(ResourceModule):
                 self.addcmd(
                     {
                         "sequence": hseq.get("sequence", None),
+                        "afi": afi,
                     },
                     "remarks_no_data",
                     negate=True,
--- a/plugins/module_utils/network/ios/rm_templates/acls.py
+++ b/plugins/module_utils/network/ios/rm_templates/acls.py
@@ -27,7 +27,10 @@ def remarks_with_sequence(remarks_data):
     if remarks_data.get("remarks"):
         cmd += " " + remarks_data.get("remarks")
     if remarks_data.get("sequence"):
-        cmd = to_text(remarks_data.get("sequence")) + " " + cmd
+        if remarks_data.get("afi") == "ipv6":
+            cmd = "sequence " + to_text(remarks_data.get("sequence")) + " " + cmd
+        else:
+            cmd = to_text(remarks_data.get("sequence")) + " " + cmd
     return cmd
 
 
@@ -258,7 +261,7 @@ class AclsTemplate(NetworkTemplate):
                     $""",
                 re.VERBOSE,
             ),
-            "setval": "remark {{ remarks }}",
+            "setval": remarks_with_sequence,
             "result": {
                 "acls": {
                     "{{ acl_name|d() }}": {
--- a/tests/integration/targets/ios_acls/tests/cli/remarks_states.yaml
+++ b/tests/integration/targets/ios_acls/tests/cli/remarks_states.yaml
@@ -130,5 +130,91 @@
     - ansible.builtin.assert:
         that:
           - base_config_overridden.commands|symmetric_difference(remarks_check_override.commands) == []
+
+    - name: Apply IPv6 ACL configuration with remarks (merged)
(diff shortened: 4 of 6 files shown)


TheNova22 (changes requested):
There currently exists an [integration test file](https://github.com/ansible-collections/cisco.ios/blob/main/tests/integration/targets/ios_acls/tests/cli/remarks_states.yaml) for this covering only IPv4, could you please update the same and have IPv6 with remarks included in it.

TheNova22 (approved):

PR #1345, opened 2026-06-08
fix(ios_bgp_address_family): add vpls as valid safi for l2vpn

### LINKED ISSUE #1202: Cisco BGP l2vpn address-family issue
##### SUMMARY

The module ios_bgp_address_family returns error when trying to create l2vpn address family. 
If "safi" option is specified as "evpn" the play works. With any other option or none at all it returns error:
"msg": "value of safi must be one of: flowspec, mdt, multicast, mvpn, evpn, unicast, got: vpls found in config -> address_family"
While playbook returns error the config is written to device fine without "safi" option or option value of "vpls".
Like this:
router bgp 12345
 bgp router-id 192.168.1.1
 bgp log-neighbor-changes
 bgp graceful-restart
 no bgp default ipv4-unicast
 !
 address-family ipv4
 exit-address-family
 !
 address-family l2vpn vpls
  neighbor 192.168.2.1 activate
 exit-address-family

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

module > ios_bgp_address_family

##### ANSIBLE VERSION

```paste below
ansible [core 2.18.6]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/<user>/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/<user>/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.12.3 (main, Sep 11 2024, 14:17:37) [GCC 13.2.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
Collection Version
---------- -------
cisco.ios  10.1.0 
```

##### CONFIGURATION

```paste below
CONFIG_FILE() = /etc/ansible/ansible.cfg
DEFAULT_HOST_LIST(/etc/ansible/ansible.cfg) = ['/home/<user>/Ansible']

GALAXY_SERVERS:
```

##### OS / ENVIRONMENT

System image file is "bootflash:asr920-universalk9_npe.17.09.07.SPA.bin"
cisco ASR-920-24SZ-M (Freescale P2020) processor (revision 1.2 GHz) with 853132K/6147K bytes of memory.

##### STEPS TO REPRODUCE

Running below playbook against inventory of cisco ios hosts.

```yaml
- name: Configure BGP vpls neighbors
  hosts: cisco_hosts
  gather_facts: false
  connection: network_cli

  vars:
    local_asn: 12345
    peer_ip: 192.168.2.1

  tasks:
    - name: Configure BGP vpnv4 neighbors
        cisco.ios.ios_bgp_address_family:
          config: 
            as_number: "{{ local_asn }}"
            address_family:
              - afi: l2vpn
                safi: vpls
                neighbors:
                - activate: true
                  neighbor_address: "{{ peer_ip }}"
          state: merged
        register: bgp_vpnv4_neighbors_config_output
```

##### EXPECTED RESULTS

Address family is created and playbook returns either "ok" or "changed"

##### ACTUAL RESULTS

The playbook ran into error on task "Configure BGP vpnv4 neighbors" although the task result on device was as expected.

```paste below
"msg": "value of safi must be one of: flowspec, mdt, multicast, mvpn, evpn, unicast, got: vpls found in config -> address_family"
```

hybridx:
Closes ansible-collections/cisco.ios#1202

> **Note:** This contribution was developed with AI assistance (Cursor IDE). 
 - Note added as per -> https://docs.ansible.com/projects/ansible/latest/community/ai_policy.html
> All changes have been reviewed and validated with `ansible-test units --docker`.

## Description

- **What is being changed?** Added `vpls` as a valid `safi` choice for the `ios_bgp_address_family` module's address-family configuration.
- **Why is this change needed?** The module rejects `safi: vpls` with a validation error (`value of safi must be one of: flowspec, mdt, multicast, mvpn, evpn, unicast, got: vpls`), even though the parser regex already recognizes `vpls` and the config is successfully applied to the device.
- **How does this change address the issue?** Adds `vpls` to the `safi` choices list in both the module docstring and the argspec, aligning input validation with the parser and actual IOS device behavior.

## Type of Change

- [ ] Documentation update
- [x] Test update
- [ ] Collection release
- [ ] CI maintenance
- [ ] Workflow maintenance
- [ ] Configuration change

## Related Issue

- Fixes #1202

## Component Name

ios_bgp_address_family

## Self-Review Checklist

- [ ] I have verified the changes work with the target IOS version(s)

### Prerequisites

- IOS Version: Any IOS/IOS-XE with L2VPN VPLS support (e.g., ASR-920 with IOS-XE 17.x)
- Hardware Platform (if applicable): N/A (unit tests only; no device required)
- Test Topology: N/A

### Steps to Test

1. Set up the collection namespace directory:
   ```bash
   mkdir -p /tmp/test-collections/ansible_collections/cisco/ios
   rsync -a --exclude='.git' --exclude='.venv' ~/ansibleProjects/cisco.ios/ \
     /tmp/test-collections/ansible_collections/cisco/ios/
   ansible-galaxy collection install ansible.netcommon ansible.utils -p /tmp/test-collections

### DIFF
--- a/tests/integration/targets/ios_bgp_address_family/tests/cli/_populate_config.yaml
+++ b/tests/integration/targets/ios_bgp_address_family/tests/cli/_populate_config.yaml
@@ -58,6 +58,11 @@
           table_map:
             name: test_tableMap
             filter: true
+        - afi: l2vpn
+          safi: vpls
+          neighbors:
+            - neighbor_address: "10.1.1.1"
+              activate: true
         - afi: ipv4
           safi: mdt
           bgp:
--- a/tests/integration/targets/ios_bgp_address_family/tests/cli/_populate_config.yaml
+++ b/tests/integration/targets/ios_bgp_address_family/tests/cli/_populate_config.yaml
@@ -16,6 +16,12 @@
       - neighbor 10.1.1.2 remote-as 200
       - neighbor 10.1.1.3 remote-as 200
 
+# Probes device capability and sets the `l2vpn_vpls_supported` fact. Some
+# platforms/images used in CI don't have the license required for l2vpn
+# vpls (see _check_l2vpn_vpls_support.yaml), so that piece of the baseline
+# is populated separately below and skipped gracefully when unsupported.
+- ansible.builtin.include_tasks: _check_l2vpn_vpls_support.yaml
+
 - name: Populate BGP address family configuration
   cisco.ios.ios_bgp_address_family:
     config:
@@ -58,11 +64,6 @@
           table_map:
             name: test_tableMap
             filter: true
-        - afi: l2vpn
-          safi: vpls
-          neighbors:
-            - neighbor_address: "10.1.1.1"
-              activate: true
         - afi: ipv4
           safi: mdt
           bgp:
@@ -119,4 +120,21 @@
                 metric: 10
                 route_map: bar
 
+    state: merged
+
+# l2vpn vpls requires a platform license (e.g. 'ax' on CSR1000v or
+# 'network-advantage addon dna-advantage' on Catalyst 8000v/9000). Kept as
+# its own call so it can be skipped independently of the rest of the
+# baseline on devices/images that lack the license.
+- name: Populate l2vpn vpls address family (requires platform license)
+  when: l2vpn_vpls_supported
+  cisco.ios.ios_bgp_address_family:
+    config:
+      as_number: "65000"
+      address_family:
+        - afi: l2vpn
+          safi: vpls
+          neighbors:
+            - neighbor_address: "10.1.1.1"
+              activate: true
     state: merged
--- a/changelogs/fragments/fix_bgp_af_vpls_safi.yaml
+++ b/changelogs/fragments/fix_bgp_af_vpls_safi.yaml
@@ -1,0 +1,4 @@
+---
+bugfixes:
+  - ios_bgp_address_family - Add ``vpls`` as a valid ``safi`` choice for the
+    ``l2vpn`` address family configuration.
--- a/docs/cisco.ios.ios_bgp_address_family_module.rst
+++ b/docs/cisco.ios.ios_bgp_address_family_module.rst
@@ -7502,6 +7502,7 @@ Parameters
                                     <li>mvpn</li>
                                     <li>evpn</li>
                                     <li>unicast</li>
+                                    <li>vpls</li>
                         </ul>
                 </td>
                 <td>
--- a/plugins/module_utils/network/ios/argspec/bgp_address_family/bgp_address_family.py
+++ b/plugins/module_utils/network/ios/argspec/bgp_address_family/bgp_address_family.py
@@ -54,7 +54,15 @@ class Bgp_address_familyArgs(object):  # pylint: disable=R0903
                         },
                         "safi": {
                             "type": "str",
-                            "choices": ["flowspec", "mdt", "multicast", "mvpn", "evpn", "unicast", "vpls"],
+                            "choices": [
+                                "flowspec",
+                                "mdt",
+                                "multicast",
+                                "mvpn",
+                                "evpn",
+                                "unicast",
+                                "vpls",
+                            ],
                         },
                         "vrf": {"type": "str"},
                         "aggregate_addresses": {
--- a/plugins/module_utils/network/ios/argspec/bgp_address_family/bgp_address_family.py
+++ b/plugins/module_utils/network/ios/argspec/bgp_address_family/bgp_address_family.py
@@ -54,7 +54,7 @@ class Bgp_address_familyArgs(object):  # pylint: disable=R0903
                         },
                         "safi": {
                             "type": "str",
-                            "choices": ["flowspec", "mdt", "multicast", "mvpn", "evpn", "unicast"],
+                            "choices": ["flowspec", "mdt", "multicast", "mvpn", "evpn", "unicast", "vpls"],
                         },
                         "vrf": {"type": "str"},
                         "aggregate_addresses": {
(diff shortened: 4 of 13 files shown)


hybridx (commented):
self review done

TheNova22:
Hey @hybridx 
Thanks for looking into this and the contribution! 
Could you please add equivalent integration tests as well so that it can be tested against devices.

hybridx (commented):
@TheNova22 I've added the changes, LMK if anything else is required

tests/integration/targets/ios_bgp_address_family/tests/cli/_populate_config.yaml
```diff
@@ -58,6 +58,11 @@
           table_map:
             name: test_tableMap
             filter: true
+        - afi: l2vpn
```
TheNova22:
```
19 - name: Populate BGP address family configuration
     ^ column 3

fatal: [REDACTED]: FAILED! => {
    "changed": false,
    "msg": "Task failed: b'address-family l2vpn vpls\\r\\n% BGP: Error initializing topology\\r\\ncisco(config-router)#'"
}
```
Following entry fails to get initialized

TheNova22: Upon manual testing, this is because it requires a separate enablement command: `license boot level network-advantage addon dna-advantage`. So since it has license addon, we'll have to pick the right way to enable this.

TheNova22 (changes requested):
Additionally, upon manual testing, the integration also fails idempotency. So that also has to be evaluated. 
It ran the same command both times: `["router bgp 65000", "address-family l2vpn vpls", "neighbor 10.1.1.1 activate", "exit-address-family"]`

tests/integration/targets/ios_bgp_address_family/tests/cli/_populate_config.yaml
hybridx: Investigated both issues and verified on a CSR1000v (IOS-XE 17.3.8a, license boot level ax):

1. `% BGP: Error initializing topology`

This occurs when the device lacks the required license for l2vpn vpls. The license command is platform-specific (`license boot level ax` on CSR1000v, `license boot level network-advantage addon dna-advantage` on Cat 8000v/9000). Added a `run_commands override` that catches this specific `AnsibleConnectionFailure` and returns a clear `fail_json` with actionable guidance on which license to enable.

2. Idempotency failure

Root cause: IOS-XE silently drops `neighbor activate` inside an address-family if the neighbor isn't first defined globally under `router bgp` (e.g. `neighbor 10.1.1.1 remote-as 100`). The command is accepted without error but doesn't persist in `show running-config`, so the module re-sends it every run.

`_populate_config.yaml` already defines global neighbors (used by overridden/deleted/replaced/gathered tests), but `merged.yaml` skipped that step. Added global BGP neighbor setup in `merged.yaml` before the address-family block. Idempotency now passes: `changed: false, commands: []`.

tests/integration/targets/ios_bgp_address_family/tests/cli/_populate_config.yaml
TheNova22, line 1: Great work! 
Integration test still fails, but as updated with latest changes, with the following line:
```   
"msg": "The device rejected an address-family with '%% BGP: Error initializing topology'. This typically means the required platform license is not active (e.g. 'license boot level ax' on CSR1000v, or 'license boot level network-advantage addon dna-advantage' on Catalyst 8000v/9000). Enable the appropriate license and reload the device."
```

The CML setup provisioned will not by default have the license enabled, so you'll have to add it via ios_config or any other way before the vpls tasks run.

TheNova22 (changes requested):

tests/integration/targets/ios_bgp_address_family/tests/cli/_populate_config.yaml
hybridx: I agree but the license can't be enabled mid-test since it requires a reload. Is it okay if I add a capability checks at the start of the test that probes address-family `l2vpn` `vpls` and skips the vpls-specific tasks if the device doesn't support it? 
OR
If we want CI to always exercise the vpls path, the license would need to be permanently enabled once in the CML golden image/template since enabling it requires license boot level ax (or network-advantage addon dna-advantage) followed by a reload, that's not something we can do per-test-run without adding several minutes to every CI run. 

Until then, this PR makes the test suite at least aware of the capability.

TheNova22: True, I forgot that reload is a dependency in this case which will make integration complicated.

I believe we can always go back to _skips the vpls-specific tasks if the device doesn't support it_, but do you see any problems with reload mid run?

Sample playbook, i believe you MIGHT have to save the config before running it.
```yaml
    - name: Send reload command and confirm prompt
      ansible.netcommon.cli_command:
        command: reload
        prompt: 'Proceed with reload?'
        answer: "\r"
      register: reload_output

    - name: Reset local connection cache
      meta: reset_connection

    - name: Wait for device to come back online
      wait_for_connection:
        delay: 60
        timeout: 300
```

**Okay, no**
Since we use DHCP, I dont think the reload makes sense as IP might change mid run. 
Hence we fall back to check based approach, I am not sure about if we can enforce license on boot at image level. Lets discuss on this with the team and opt an approach accordingly

hybridx: All non-hardware CI (unit tests, Sonar, Codecov, pre-commit) is green. The only outstanding check is the CML integration run, which should now exercise the licensed path where available otherwise skip tests cleanly IMO.

TheNova22: Great work! Integration tests are now green with the check addition

TheNova22 (approved):

PR #1353, opened 2026-07-07
fix(ios_hsrp_interfaces): add compval use_bia.set to fix set:false not negating use-bia

TheNova22:
## Description

- **What**: Added `compval: "use_bia.set"` to the `use_bia` parser in `rm_templates/hsrp_interfaces.py`.
- **Why**: Without `compval`, `RMEngineBase.compare()` resolved the comparison path from `name: "use_bia"`, extracting `want["use_bia"] = {"set": False}` — a non-empty dict, which is truthy. As a result, `compare()` fired `setval` instead of `remval` when `state: merged` was used with `use_bia: {set: false}` against a device that already had `standby use-bia` configured, emitting `standby use-bia scope interface` instead of `no standby use-bia`.
- **How**: `compval: "use_bia.set"` causes `compare()` to evaluate the boolean leaf directly. `False` → `remval` fires (`no standby use-bia`); `True` with no have → `setval` fires (`standby use-bia`); `True` with matching have → no change (idempotent). Note: `deleted`/`replaced`/`overridden` states were unaffected as they bypass `compare()` and use the direct delete path.

## Type of Change

- [x] Test update

## Related Issue

- Related to #

## Component Name

`ios_hsrp_interfaces`

## Self-Review Checklist

### Prerequisites

- IOS Version: IOS-XE (CSR1000v)
- Hardware Platform (if applicable): CSR1000v
- Test Topology: Single router with GigabitEthernet3

### Steps to Test

1. Configure `standby use-bia` on GigabitEthernet3 manually or via `ios_config`
2. Run `ios_hsrp_interfaces` with `use_bia: {set: false}` and `state: merged`
3. Verify `no standby use-bia` is in the generated commands and `changed: true`
4. Run the same task again — verify `changed: false` (idempotent)
5. With no `standby use-bia` on device, run `use_bia: {set: true}` with `state: merged`
6. Verify `standby use-bia` is generated and `changed: true`; second run `changed: false`

### Expected Results

`use_bia: {set: false}` with `standby use-bia` present → `no standby use-bia`, idempotent on second run.
`use_bia: {set: true}` with no `standby use-bia` → `standby use-bia`, idempotent on second run.

### Test Results

```
# unit tests
tox -e unit-py3.11-2.19: 545 passed

# sanity
tox -e sanity-py3.11-2.18: OK

# device run (before fix)
TASK [Verify fix]
commands:
  - interface GigabitEthernet3
  - standby use-bia scope interface   # wrong: setval fired instead of remval

# device run (after fix)
TASK [Verify fix]
commands:
  - interface GigabitEthernet3
  - no standby use-bia                # correct
# second run: changed=false (idempotent confirmed)
```

- [x] All acceptance criteria have been reviewed and verified
- [x] Acceptance criteria checklist:
  - [x] `state: merged` with `use_bia: {set: false}` emits `no standby use-bia` when device has the feature configured
  - [x] `state: merged` with `use_bia: {set: true}` emits `standby use-bia` when device does not have it
  - [x] Both paths are idempotent on the second run

## Additional Context

The bug only affected `state: merged`. Other states (`deleted`, `replaced`, `overridden`) were working correctly because they bypass `compare()` and use the direct delete path when negating `use_bia`.

### Command Output / Logs

```
# Before fix — device had "standby use-bia", user requested use_bia: {set: false}
commands:
  - interface GigabitEthernet3
  - standby use-bia scope interface

# After fix
commands:
  - interface GigabitEthernet3
  - no standby use-bia
```

### Required Actions

- [x] Requires changelog fragment
- [x] Requires integration test updates
- [x] Requires unit test updates


malikdotexe (approved):
Looks Good

TheNova22:
Note: The following is being considered as compval addition because of the nature of the parameter present. Ideally the change would be to replace the parameter and make use_bia as a bool, but that will classify as breaking_change.
Using comp_val here hence makes it easier because of how this param was designed and we avoid re-implementing what compval was designed to do as a function in parser comparision.

PR #1359, opened 2026-07-17
Fix ios_user aggregation issue with update_password & hashed_password idempotency

### LINKED ISSUE #559: ios_user: update_password key failure with aggregate. + others
##### SUMMARY

I noticed 3 different issues with this module:

1) module breaks when the update_password is used in the aggregate dictionary.

```
  cisco.ios.ios_user:
    aggregate:
      - name: "{{ username1 }}"
        configured_password: "{{ clear passwd1 }}" 
        update_password: 'always'
      - name: "{{ username2 }}"
        configured_password: "{{ clear passwd2 }}" 
        update_password: 'always'
```
(this breaks)

works as expected is defined like this : 

```
  cisco.ios.ios_user:
    aggregate:
      - name: "{{ username1 }}"
        configured_password: "{{ clear passwd1 }}" 
      - name: "{{ username2 }}"
        configured_password: "{{ clear passwd2 }}" 
      update_password: 'always'
```
(this works)

2) some hashes are not blurred in the output with no log

expected result : 
        "username dummy secret ******** ********"
actual result:
        "username dummy secret ******** $********$human_readable_hash********",

This behaviour depends on the hash. Some are blurred others are clear.

3) ( more of an enhancement ) 
There is no choice regarding encryption.
The only algorithm available through the module is md5 which considered broken.
The only recommanded encryption algorithm, as of today, is scrypt.
Even if it is possible to configure the device with a hashed password ... it will set the same hash on all configuration files which is not desirable neither.

Proposition: add a new key word in the module to choose the encryption algorithm with clear password as input.
the command to be :
`username <username> privilege 15 algorithm-type scrypt secret <secret>`

### Work around:

Use the module "naked" with purge key set to true;
Configure desired users and passwords with config module in another task.

##### ISSUE TYPE
- Bug Report

##### ANSIBLE VERSION

```paste below
github latestansible [core 2.12.2]
  config file = /remote/users/mcarrouee/.ansible.cfg
...
  python version = 3.8.8 (default, Aug 11 2021, 06:52:42) [GCC 8.5.0 20210514 (Red Hat 8.5.0-3)]
  jinja version = 3.0.3
  libyaml = True
```

##### COLLECTION VERSION

```paste below
github latest
```

##### CONFIGURATION

```paste below

```

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

module failure with aggregate

```paste below
The full traceback is:
Traceback (most recent call last):
  File "~/.ansible/tmp/ansible-local-2972609kf089_dt/ansible-tmp-1649944005.6931295-2972638-226093719612968/AnsiballZ_ios_user.py", line 107, in <module>
    _ansiballz_main()
  File "~/.ansible/tmp/ansible-local-2972609kf089_dt/ansible-tmp-1649944005.6931295-2972638-226093719612968/AnsiballZ_ios_user.py", line 99, in _ansiballz_main
    invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
  File "~/.ansible/tmp/ansible-local-2972609kf089_dt/ansible-tmp-1649944005.6931295-2972638-226093719612968/AnsiballZ_ios_user.py", line 47, in invoke_module
    runpy.run_module(mod_name='ansible_collections.cisco.ios.plugins.modules.ios_user', init_globals=dict(_module_fqn='ansible_collections.cisco.ios.plugins.modules.ios_user', _modlib_path=modlib_path),
  File "/usr/lib64/python3.8/runpy.py", line 207, in run_module
    return _run_module_code(code, init_globals, run_name, mod_spec)
  File "/usr/lib64/python3.8/runpy.py", line 97, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/usr/lib64/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/tmp/ansible_cisco.ios.ios_user_payload_1zvvkg4k/ansible_cisco.ios.ios_user_payload.zip/ansible_collections/cisco/ios/plugins/modules/ios_user.py", line 621, in <module>
  File "/tmp/ansible_cisco.ios.ios_user_payload_1zvvkg4k/ansible_cisco.ios.ios_user_payload.zip/ansible_collections/cisco/ios/plugins/modules/ios_user.py", line 605, in main
  File "/tmp/ansible_cisco.ios.ios_user_payload_1zvvkg4k/ansible_cisco.ios.ios_user_payload.zip/ansible_collections/cisco/ios/plugins/modules/ios_user.py", line 550, in update_objects
KeyError: 'update_password'
```

TheNova22:
## Description

- What is being changed?
  
  - update_password and password_type are now resolved per aggregate item using get_param_value, allowing each entry in an aggregate list to independently override the module-level defaults.
  - A new parse_hashed_password() helper parses the existing hash type and value from running config, so repeated runs with the same hashed_password type/value pair correctly produce no commands.
  - Documentation for update_password clarifies that on_create/always applies only to configured_password; hashed_password always uses diff-based comparison.

- Why is this change needed?

  - `map_obj_to_commands` read update_password and password_type from module.params once at function entry. Per-item overrides set inside aggregate entries were silently ignored — all items shared the module-level value. If update_password was set only inside an aggregate item (not at the task level), the module-level default (always) took effect instead, causing password commands to be re-sent for existing users on every run.
  - `map_config_to_obj` always initialised hashed_password to None. Because have["hashed_password"] was always None, needs_update() treated every run as a change when want["hashed_password"] was set — re-applying the same hash type and value on every invocation.

- How does this change address the issue?

  - Per-item update_password and password_type are now stored in each want dict during map_params_to_obj via get_param_value, which handles the aggregate-level → task-level fallback automatically.
  - map_obj_to_commands reads them from want inside the per-item loop instead of from module.params before the loop.
  - parse_hashed_password() extracts {type, value} from the running config line, enabling a true diff against want["hashed_password"].

## Type of Change

- [x] Documentation update
- [ ] Test update
- [ ] Collection release
- [ ] CI maintenance
- [ ] Workflow maintenance
- [ ] Configuration change

## Related Issue

- Fixes #559 

## Component Name

ios_user

## Self-Review Checklist

Assisted By: Claude

### DIFF
--- a/changelogs/fragments/ios_user_aggregate_password_fixes.yaml
+++ b/changelogs/fragments/ios_user_aggregate_password_fixes.yaml
@@ -10,8 +10,6 @@ bugfixes:
   - ios_user - fixed hashed_password idempotency so that re-applying the same
     type/value pair against an already-configured user produces no commands,
     preventing unnecessary password updates on repeat runs.
-
-minor_changes:
   - ios_user - update_password and password_type are now resolved per aggregate
     item via get_param_value, allowing each entry in the aggregate list to
     independently override the module-level defaults
--- a/changelogs/fragments/ios_user_aggregate_password_fixes.yaml
+++ b/changelogs/fragments/ios_user_aggregate_password_fixes.yaml
@@ -1,4 +1,11 @@
 ---
+doc_changes:
+  - ios_user - clarified that update_password (on_create/always) applies only to
+    configured_password; hashed_password always uses hash type and value comparison
+    to determine whether a change is required.
+  - ios_user - documented type 9 (scrypt) as a valid hash type alongside the
+    existing type 5 (MD5) and type 8 (PBKDF2) examples for hashed_password.
+
 bugfixes:
   - ios_user - fixed hashed_password idempotency so that re-applying the same
     type/value pair against an already-configured user produces no commands,
--- a/changelogs/fragments/ios_user_aggregate_password_fixes.yaml
+++ b/changelogs/fragments/ios_user_aggregate_password_fixes.yaml
@@ -1,7 +1,5 @@
 ---
 bugfixes:
-  - ios_user - removed accidental debugpy breakpoint import from main() that
-    blocked module execution waiting for a debugger connection.
   - ios_user - fixed hashed_password idempotency so that re-applying the same
     type/value pair against an already-configured user produces no commands,
     preventing unnecessary password updates on repeat runs.
@@ -9,8 +7,7 @@ bugfixes:
 minor_changes:
   - ios_user - update_password and password_type are now resolved per aggregate
     item via get_param_value, allowing each entry in the aggregate list to
-    independently override the module-level defaults rather than all items
-    sharing a single value read from module.params in the command-generation loop.
-  - ios_user - parse_hashed_password and parse_password_type helpers now extract
-    the stored hash type, hash value, and keyword (secret/password) from running
-    config, enabling proper diff-based idempotency checks for hashed_password.
+    independently override the module-level defaults
+  - ios_user - parse_hashed_password  helper now extracts the stored hash type,
+    hash value from running config, enabling proper diff-based idempotency
+    checks for hashed_password.
--- a/changelogs/fragments/ios_user_aggregate_password_fixes.yaml
+++ b/changelogs/fragments/ios_user_aggregate_password_fixes.yaml
@@ -1,0 +1,16 @@
+---
+bugfixes:
+  - ios_user - removed accidental debugpy breakpoint import from main() that
+    blocked module execution waiting for a debugger connection.
+  - ios_user - fixed hashed_password idempotency so that re-applying the same
+    type/value pair against an already-configured user produces no commands,
+    preventing unnecessary password updates on repeat runs.
+
+minor_changes:
+  - ios_user - update_password and password_type are now resolved per aggregate
+    item via get_param_value, allowing each entry in the aggregate list to
+    independently override the module-level defaults rather than all items
+    sharing a single value read from module.params in the command-generation loop.
+  - ios_user - parse_hashed_password and parse_password_type helpers now extract
+    the stored hash type, hash value, and keyword (secret/password) from running
+    config, enabling proper diff-based idempotency checks for hashed_password.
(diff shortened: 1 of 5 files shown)


komaldesai13:
check

changelogs/fragments/ios_user_aggregate_password_fixes.yaml
```diff
@@ -0,0 +1,20 @@
...
+    to determine whether a change is required.
+  - ios_user - documented type 9 (scrypt) as a valid hash type alongside the
+    existing type 5 (MD5) and type 8 (PBKDF2) examples for hashed_password.
+
+bugfixes:
+  - ios_user - fixed hashed_password idempotency so that re-applying the same
+    type/value pair against an already-configured user produces no commands,
+    preventing unnecessary password updates on repeat runs.
+
+minor_changes:
```
komaldesai13: Update_password suppose to work per item in task like other sub-options under aggregate , which was not working , and hence we are fixing it , i would consider it as bugfix only

plugins/modules/ios_user.py
```diff
@@ -64,7 +64,9 @@
           - Since passwords are encrypted in the device running config, this argument will
             instruct the module when to change the password.  When set to C(always), the
             password will always be updated in the device and when set to C(on_create) the
-            password will be updated only if the username is created.
+            password will be updated only if the username is created. This is applicable for
+            C(configured_password), while C(hashed_password) will undergo hash and type
+            comparision to check whether a change is required.
         choices:
           - on_create
```
komaldesai13, line 71: As we are mentioning this ,can we add some test for negative scenario here ? integration or unit ?

changelogs/fragments/ios_user_aggregate_password_fixes.yaml
TheNova22: Done! The change is now bugfixes + doc_changes only. Thanks!

plugins/modules/ios_user.py
TheNova22: Added two new units about the statement mentioned:

- test_aggregate_hashed_password_on_create_does_not_suppress_hash_comparison
- test_aggregate_hashed_password_type_and_value_mismatch

komaldesai13 (approved):

PR #1369, opened 2026-08-16
fix(ios_bgp_global): accept ASDOT local_as notation (type int -> str for local_as.number)

TheNova22:
## Description
- **What:** Changes `local_as.number` argspec type from `int` to `str` in `plugins/module_utils/network/ios/argspec/bgp_global/bgp_global.py`.
- **Why:** When a device has `bgp asnotation dot` enabled and a neighbor configured with `local-as 501.65083` (ASDOT 4-byte AS notation), `ast.literal_eval("501.65083")` produces a Python float. Ansible's `check_type_int()` rejects it, crashing `ios_bgp_global` and `ios_facts` with _"argument 'number' is of type float and we were unable to convert to int"_.
- **How:** Type `local_as.number` as `str` — consistent with `remote_as` in the same argspec which already accepts ASDOT. Ansible's `check_type_str()` silently coerces plain integers to strings, so existing playbooks passing `local_as.number: 234` (int) are **not broken**. The setval Jinja2 template already uses `local_as.number|string`, so no `rm_templates` change is needed.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Test update

## Related Issue
- Fixes #1096

## Component Name
ios_bgp_global

## Self-Review Checklist
- [x] I have performed a self-review of my code
- [x] I have removed all commented code (no commented code should be merged)
- [x] I have considered the security impact of these changes
- [x] I have considered performance implications
- [x] I have thought about error handling and edge cases
- [x] I have tested the changes in my local environment
- [x] I have verified the changes work with the target IOS version(s)
- [x] I have reviewed the acceptance criteria for related tickets

### Prerequisites
- IOS-XE device with `bgp asnotation dot` configured globally

### Steps to Test
1. Configure: `router bgp 65000 / bgp asnotation dot / neighbor 192.0.2.1 remote-as 100 / neighbor 192.0.2.1 local-as 501.65083`
2. Run `ios_bgp_global state=gathered` — verify `local_as.number` is string `"501.65083"`, no type error
3. Run `ios_bgp_global state=merged` with `local_as: {number: "501.65083"}` — verify `changed: false`
4. Backward compat: run merged with `local_as: {number: 234}` (plain int) — must not fail
5. Combined: configure ASDOT for both `remote-as` and `local-as`; verify both gathered as strings

### Expected Results
- Steps 2–3: No errors; ASDOT local_as.number gathered as string; merged idempotent
- Step 4: Integer `234` coerced to `"234"` — no breaking change
[...]

PR #1371, opened 2026-08-19
fix(ios_bgp_address_family): fix idempotency when remote_as is specified for a neighbor

### LINKED ISSUE #1190: Bug Report: cisco.ios.ios_bgp_address_family Idempotency Issues with state: merged / replaced / overridden
##### SUMMARY
The cisco.ios.ios_bgp_address_family module exhibits idempotency issues when using state: replaced. The module repeatedly applies configuration changes even when the current state matches the desired state. This behavior leads to unnecessary reconfiguration and potential disruption in production environments.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
cisco.ios.ios_bgp_address_family

##### ANSIBLE VERSION
ansible [core 2.15.0]
config file = /etc/ansible/ansible.cfg
configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
ansible python module location = /usr/lib/python3.8/site-packages/ansible
python version = 3.8.10
jinja version = 3.1.2
libyaml = True

##### COLLECTION VERSION
cisco.ios 4.2.0

##### CONFIGURATION
DEFAULT_HOST_LIST(/etc/ansible/ansible.cfg) = ['/etc/ansible/hosts']
DEFAULT_TIMEOUT(/etc/ansible/ansible.cfg) = 30

```

##### OS / ENVIRONMENT
Target Device: Cisco Catalyst 9300
IOS Version: 17.3.3
Control Node OS: Ubuntu 22.04


##### STEPS TO REPRODUCE
1. Use the cisco.ios.ios_bgp_address_family module to configure a BGP neighbor and networks.
2. Run the playbook multiple times with the same configuration.
3. Observe that the module repeatedly applies configuration changes even when the current state matches the desired state.

---
- hosts: group1
  gather_facts: no
  connection: ansible.netcommon.network_cli
  vars:
    ansible_network_os: "ios"

  tasks:
  - name: Manage BGP address family config
    cisco.ios.ios_bgp_address_family:
      config:
        as_number: 65555
        address_family:
        - afi: ipv4
          networks:
            - address: 192.0.2.3
            - address: 192.0.2.2
          neighbors:
            - neighbor_address: 192.0.2.1
              description: test1
              remote_as: 100
              route_maps:
                - name: test_rm
                  out: true
            - neighbor_address: 192.0.3.1
              description: test2
              remote_as: 100
              route_maps:
                - name: test_rm
                  in: true
      state: replaced


<!--- Paste example playbooks or commands between quotes below -->


##### EXPECTED RESULTS
The module should detect that the current configuration matches the desired state and report no changes (changed: false).


##### ACTUAL RESULTS
The module repeatedly applies configuration changes, resulting in unnecessary commands being issued to the device.

```
TASK [Manage BGP address family config] ********************************************************************************************************************************************************************************************
changed: [lab_9300] => {
    "changed": true,
    "commands": [
        "router bgp 65555",
        "address-family ipv4",
        "neighbor 192.0.2.1 remote-as 100",
        "no neighbor 192.0.2.1 activate",
        "neighbor 192.0.2.1 description test1",
        "neighbor 192.0.2.1 route-map test_rm out",
        "neighbor 192.0.3.1 remote-as 100",
        "no neighbor 192.0.3.1 activate",
        "neighbor 192.0.3.1 description test2",
        "neighbor 192.0.3.1 route-map test_rm in",
        "network 192.0.2.2"
    ]
}

```

Analysis

Root Cause

The idempotency issue arises due to incomplete or flawed comparison logic in the module. Specifically:

Nested Attributes:
Attributes like route_maps, description, and activate are not being compared deeply, causing the module to interpret the current state as different from the desired state.

Default Values:
Null or default values for attributes are not being handled correctly, leading to unnecessary changes.
State Enforcement:

The state: replaced option should enforce strict idempotency, but the module applies changes incrementally instead of enforcing a strict match.
Please note that I also tried merge and overridden, and in all cases it fails.

Impact
This bug causes:

Unnecessary Configuration Changes:
Commands are repeatedly issued even when the current state matches the desired state.
Disruption in Production Environments:
Reapplying configuration can cause temporary disruptions in BGP sessions or routing.
Loss of Idempotency:
The behavior breaks the fundamental principle of idempotency in Ansible playbooks.

Suggested Fix
Code Fix
The module's comparison logic should be updated to:

Perform a deep comparison of nested attributes like route_maps, description, and activate.
Handle null values and defaults properly to avoid interpreting the current state as incomplete.
Enforce strict idempotency for state: replaced.


`
def compare_neighbors(desired_neighbors, current_neighbors):
    commands = []
    for neighbor in desired_neighbors:
        current_neighbor = next((n for n in current_neighbors if n['neighbor_address'] == neighbor['neighbor_address']), {})
        
        # Compare 'activate'
        if neighbor.get('activate') is not None and neighbor['activate'] != current_neighbor.get('activate'):
            commands.append(f"neighbor {neighbor['neighbor_address']} {'activate' if neighbor['activate'] else 'no activate'}")
        
        # Compare 'route_maps'
        if 'route_maps' in neighbor:
            for route_map in neighbor['route_maps']:
                current_route_map = next((rm for rm in current_neighbor.get('route_maps', []) if rm['name'] == route_map['name']), None)
                if not current_route_map or current_route_map != route_map:
                    if route_map.get('in'):
                        commands.append(f"neighbor {neighbor['neighbor_address']} route-map {route_map['name']} in")
                    if route_map.get('out'):
                        commands.append(f"neighbor {neighbor['neighbor_address']} route-map {route_map['name']} out")
        
        # Compare 'description'
        if neighbor.get('description') and neighbor['description'] != current_neighbor.get('description'):
            commands.append(f"neighbor {neighbor['neighbor_address']} description {neighbor['description']}")
    
    return commands
`

### PR DESCRIPTION
**Pull request by TheNova22**
## Description

- **What is being changed?**
  `ios_bgp_address_family` re-emitted `neighbor X remote-as Y` on every run even when the config was already present on the device.

- **Why is this change needed?**
  IOS stores `neighbor X remote-as Y` at global BGP scope (1-space indent) regardless of whether it was entered inside `address-family`. The existing parser used `\s\s` (2-space AF indent), so it never matched, and `have` was always empty for `remote_as` — causing the command to be re-issued every run.

- **How does this change address the issue?**
  Two coordinated changes:
  1. **`rm_templates/bgp_address_family.py`** — regex widened from `\s\s` to `\s+` so it matches the 1-space global indent IOS uses for this line. Globally-parsed `remote-as` lines land in the internal `"__"` (no-AF) bucket.
  2. **`config/bgp_address_family.py`** — the `"__"` bucket is extracted before AF comparison. `_compare_neighbor_lists` receives these as `global_neighbors` and injects `remote_as` into `have_nbr` when want specifies it but the AF-specific have does not, restoring correct idempotency.

## Type of Change
- [x] Test update

## Related Issue
- Fixes #1190 

## Component Name
`ios_bgp_address_family`

## Self-Review Checklist

## Additional Context
IOS device behavior that motivates this fix:

```
device(config-router)#address-family ipv4
device(config-router-af)#neighbor 192.0.2.1 remote-as 100
device(config-router-af)#neighbor 192.0.2.1 activate
device(config-router-af)#exit-address-family
device(config)#do show run | section router bgp
router bgp 65000
 neighbor 192.0.2.1 remote-as 100    ← 1-space: global BGP, not AF
 address-family ipv4
  neighbor 192.0.2.1 activate        ← 2-space: AF level
 exit-address-family
```
Even when entered inside address-family, IOS promotes remote-as to global BGP scope. The old \s\s regex targeted AF-level indentation and never matched it.

`remote_as` is already a supported parameter in `ios_bgp_global`. It is also present in the `ios_bgp_address_family` - this is intentional, since IOS accepts the command inside an address-family context even though the resulting config line always appears at global BGP scope. This fix wires up the global-level lookup so the existing argspec support becomes correctly idempotent. The alternative to this will be removing remote_as from the bgp_address_family argspec entirely and requiring users to manage it exclusively via ios_bgp_global (a breaking change).

Assisted By: Claude Code

### DIFF
--- a/plugins/modules/ios_bgp_address_family.py
+++ b/plugins/modules/ios_bgp_address_family.py
@@ -799,8 +799,16 @@ options:
                     type: bool
               remote_as:
                 description:
-                  - Specify a BGP neighbor
-                  - AS of remote neighbor
+                  - Specify a BGP neighbor remote AS number.
+                  - On IOS, C(neighbor X remote-as Y) always resides at the
+                    global BGP level not inside an address-family block.
+                  - This module reads the global-level value for idempotency so
+                    that specifying C(remote_as) here does not generate a
+                    spurious C(neighbor X remote-as Y) command on every run.
+                  - C(ios_bgp_address_family) does not add, change, or remove the
+                    global C(neighbor X remote-as Y) statement. Use M(cisco.ios.ios_bgp_global)
+                    to manage that statement. The deleted and replaced states
+                    will not remove a global remote-as.
                 type: str
               remove_private_as:
                 description: Remove private AS number from outbound updates
--- a/changelogs/fragments/fix-bgp-af-remote-as-not-captured.yaml
+++ b/changelogs/fragments/fix-bgp-af-remote-as-not-captured.yaml
@@ -5,3 +5,10 @@ bugfixes:
     address-family and at global level, but always resides at the global level. Hence, the
     config class now correctly associates global-level attributes with the corresponding
     address-family during have-facts comparison.
+doc_changes:
+  - >-
+    ios_bgp_address_family - Document the ``remote_as`` lifecycle gap. On IOS,
+    ``neighbor X remote-as Y`` lives at the global BGP level, not inside an address-family
+    block. This module reads the global value for idempotency but does not add, change, or
+    remove it; use ``ios_bgp_global`` to manage the global remote-as statement.
+    The ``deleted`` and ``replaced`` states will not remove a global remote-as entry.
--- a/changelogs/fragments/fix-bgp-af-remote-as-not-captured.yaml
+++ b/changelogs/fragments/fix-bgp-af-remote-as-not-captured.yaml
@@ -1,0 +1,7 @@
+---
+bugfixes:
+  - ios_bgp_address_family - Fix idempotency issue where specifying ``remote_as`` for a neighbor
+    caused the module to emit ``neighbor X remote-as Y`` on every run.  It can be specified at
+    address-family and at global level, but always resides at the global level. Hence, the
+    config class now correctly associates global-level attributes with the corresponding
+    address-family during have-facts comparison.
--- a/docs/cisco.ios.ios_bgp_address_family_module.rst
+++ b/docs/cisco.ios.ios_bgp_address_family_module.rst
@@ -4217,8 +4217,10 @@ Parameters
                 <td>
                 </td>
                 <td>
-                        <div>Specify a BGP neighbor</div>
-                        <div>AS of remote neighbor</div>
+                        <div>Specify a BGP neighbor remote AS number.</div>
+                        <div>On IOS, <code>neighbor X remote-as Y</code> always resides at the global BGP level not inside an address-family block.</div>
+                        <div>This module reads the global-level value for idempotency so that specifying <code>remote_as</code> here does not generate a spurious <code>neighbor X remote-as Y</code> command on every run.</div>
+                        <div><code>ios_bgp_address_family</code> does not add, change, or remove the global <code>neighbor X remote-as Y</code> statement. Use <span class='module'>cisco.ios.ios_bgp_global</span> to manage that statement. The deleted and replaced states will not remove a global remote-as.</div>
                 </td>
             </tr>
             <tr>
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -14,3 +14,6 @@ tags: [cisco, ios, iosxe, networking]
 version: "11.5.1"
 build_ignore:
   - .ansible-lint
+  - .venv
+  - collections
+  - .tox
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -14,6 +14,3 @@ tags: [cisco, ios, iosxe, networking]
 version: "11.5.1"
 build_ignore:
   - .ansible-lint
-  - .venv
-  - collections
-  - .tox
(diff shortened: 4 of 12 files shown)

### CODE REVIEW

File: plugins/modules/ios_bgp_address_family.py [hunk 1]
```diff
@@ -799,8 +799,16 @@
                     type: bool
               remote_as:
                 description:
-                  - Specify a BGP neighbor
-                  - AS of remote neighbor
+                  - Specify a BGP neighbor remote AS number.
```
komaldesai13, line 802: why number is being assigned as str , can we change this to int

komaldesai13 (commented):
can we add unit/integrationtests for deleted state also to make sure it works well

plugins/modules/ios_bgp_address_family.py
TheNova22: This is `str` across different modules to support AS-DOT notation, similar to local-as raised recently as part of https://github.com/ansible-collections/cisco.ios/pull/1369 (Reference issue: https://github.com/ansible-collections/cisco.ios/issues/1096) 
```
pombon(config-router)#neighbor 192.0.2.1 remote-as 23.35  

pombon(config-router)#neighbor 192.0.2.1 remote-as ? 
  <1-4294967295>  AS of remote neighbor
  <1.0-XX.YY>     AS of remote neighbor
```

TheNova22:
> can we add unit/integrationtests for deleted state also to make sure it works well

@komaldesai13 - Done! I added an entry of `remote_as` for `test_ios_bgp_address_family_deleted`, an already existing unit test case. 
Integration test for deleted already consumes `_populate_config.yaml`, so that is being validated.
But it isn't a major check in this code change because it always returns `no address-family *` and nothing related to `remote_as`

komaldesai13 (approved):

PR #1373, opened 2026-08-25
fix(ios_route_maps): emit route-map header for bare entries (action+sequence only)

### LINKED ISSUE #1234: ios_route_maps does not apply a route map entry if it only contains a sequence number and an action
##### SUMMARY
ios_route_maps does not apply a route map entry if it only contains a sequence number and an action.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
cisco.ios.ios_route_maps
##### ANSIBLE VERSION

```paste below
2.18.5
```

##### COLLECTION VERSION

```paste below
10.1.0
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT

Running in AWX in a custom execution environment.

##### STEPS TO REPRODUCE

```yaml
      - action: deny #  This is not applied by ios_route_maps so there is a separate ios_config task to make sure it is applied
        sequence: 6

```
This same bug also ocurrs in a different route map with a permit entry.

##### EXPECTED RESULTS

If the above item is in config for the module, this entry will not appear in the commands during check mode. We did not attempt to apply it in run mode, so it's possible this bug is limited to reporting the line in commands output.

##### ACTUAL RESULTS

```paste below

```

TheNova22:
## Description

- **What is being changed?** `entries_compare()` in the `ios_route_maps` config class now correctly emits the route-map header command for entries that contain only `action` and `sequence` (no `description`, `match`, or `set` clauses).
- **Why is this change needed?** Bare entries were silently dropped. The `cmd_len` guard only inserted the route-map header when a sub-command was generated; bare entries produce no sub-commands, so `cmd_len` never changed and the header (`route-map MYMAP deny 6`) was never emitted.
- **How does this change address the issue?** The header insert is moved inside the `if want["entries"][k] != have_entry:` block (Branch 1), so it fires whenever an entry is new or changed — regardless of sub-command count. In the no-have branch (Branch 2) the stale guard is removed entirely since all entries are new. A `_route_map_entry_cmd()` helper consolidates the repeated header-building logic across all three call sites.

## Type of Change

- [x] Refactoring (no functional changes)
- [x] Test update

## Related Issue

- Fixes #1234

## Component Name

`ios_route_maps`

## Self-Review Checklist

### Prerequisites

- IOS Version: IOS XE (tested on lab device via network_cli)
- No special topology required — the bug is in command generation logic

### Steps to Test

1. Ensure no `route-map MYMAP` exists on the device
2. Run `ios_route_maps` with `state: merged` and two bare entries (`action`+`sequence` only)
3. Verify `commands` includes `route-map MYMAP deny 6` and `route-map MYMAP permit 10`
4. Run the same task again — verify `changed: false` (idempotency)

### Expected Results

Run 1: `changed: true`, `commands: ["route-map MYMAP deny 6", "route-map MYMAP permit 10"]`
Run 2: `changed: false`, `commands: []`

### Test Results

**Before fix:**
```
TASK [Apply bare route-map entries]
ok: changed=false commands=[]   ← entries silently dropped
```

**After fix:**
```
TASK [Apply bare route-map entries]
changed: changed=true commands=["route-map TESTMAP_BARE deny 6", "route-map TESTMAP_BARE permit 10"]

TASK [Idempotency check]
ok: changed=false commands=[]   ← idempotency preserved
```

Unit: `578 passed` (2 new regression tests added)
Sanity: `OK`

- [x] All acceptance criteria have been reviewed and verified
- [x] Acceptance criteria checklist:
  - [x] Both `entries_compare()` branches emit the route-map header for bare entries
  - [x] Unit test: bare entries on empty device → commands include both headers
  - [x] Unit test: bare entries already on device → `changed: false`, `commands: []`
  - [x] Changelog fragment added under `bugfixes`

### Root Cause

`entries_compare()` had a guard `if cmd_len != len(self.commands):` that acted as a proxy for "were any sub-commands generated?". For bare entries, no sub-commands are generated, so `cmd_len` stayed equal to `len(self.commands)` and the header was never inserted. The correct condition is simply whether the entry differs from what's on the device (`want != have_entry`), which was already checked by the inner `if` block — the header insert just needed to move inside it.

### Workaround (before fix)

Adding any one sub-clause worked around the bug:
```yaml
entries:
  - action: deny
    sequence: 6
    description: placeholder   # forced header to emit
```

### Command Output / Logs

```
# Before fix
result.commands = []
result.changed = false

# After fix
result.commands = ["route-map TESTMAP_BARE deny 6", "route-map TESTMAP_BARE permit 10"]
result.changed = true
```

### Required Actions

- [x] Requires changelog fragment — added `changelogs/fragments/fix-route-maps-bare-entry.yaml`
- [x] Requires integration test updates — bare-entry scenario added to `tests/integration/targets/ios_route_maps/tests/cli/merged.yaml`
- [x] Requires unit test updates — two new methods in `test_ios_route_maps.py`

### DIFF
--- a/changelogs/fragments/fix-route-maps-bare-entry.yaml
+++ b/changelogs/fragments/fix-route-maps-bare-entry.yaml
@@ -1,0 +1,6 @@
+bugfixes:
+  - ios_route_maps - fix bare entries (action+sequence only) being silently
+    dropped; the cmd_len guard in entries_compare() now emits the route-map
+    header command even when no sub-commands are generated (description/match/set
+    absent), so catch-all rules like ``route-map MYMAP deny 6`` are correctly
+    applied.
--- a/plugins/module_utils/network/ios/config/route_maps/route_maps.py
+++ b/plugins/module_utils/network/ios/config/route_maps/route_maps.py
@@ -105,6 +105,14 @@ class Route_maps(ResourceModule):
         if want != have and self.state != "deleted":
             self.entries_compare(want, have)
 
+    def _route_map_entry_cmd(self, route_map_name, entry, negate=False):
+        cmd = ("no " if negate else "") + "route-map {0}".format(route_map_name)
+        if entry.get("action"):
+            cmd += " {0}".format(entry["action"])
+        if entry.get("sequence"):
+            cmd += " {0}".format(entry["sequence"])
+        return cmd
+
     def entries_compare(self, want, have):
         if want.get("entries"):
             cmd_len = len(self.commands)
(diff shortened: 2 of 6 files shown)


malikdotexe (approved):
LGTM

PR #1377, opened 2026-09-05
fix: ios_vrf_global (state=replaced) cannot change VRF RD on IOS-

### LINKED ISSUE #1293: ios_vrf_global (state=replaced) cannot change VRF RD on IOS-XE 17.9.5a (missing "no rd <old>" / reports changed but config unchanged)
### Bug Summary

**Summary**
When managing VRFs with cisco.ios.ios_vrf_global using state: replaced, changing the RD of an existing VRF fails on IOS-XE 17.9.5a because IOS requires removing the old RD first (no rd <old>) before setting the new one. The module only sends rd <new> and reports changed: true, but before and after remain identical and the RD is not updated.

**Environment**
Ansible: 2.14.6

Collection: cisco.ios 11.2.0

Connection: network_cli

Device OS: Cisco IOS-XE 17.9.5a

Platform/Model: Cisco C8000V

Transport: SSH

**Steps to Reproduce (minimal playbook)**

```
- name: Repro - change RD with ios_vrf_global
  hosts: routers
  gather_facts: false
  connection: network_cli
  collections:
    - cisco.ios

  tasks:
    - name: Ensure VRF exists with an initial RD (only for clean repro)
      cisco.ios.ios_config:
        parents:
          - "vrf definition vlan222-vrf"
        lines:
          - "rd 10.202.200.97:222"

    - name: Attempt to change RD via ios_vrf_global (fails)
      cisco.ios.ios_vrf_global:
        config:
          vrfs:
            - name: "vlan222-vrf"
              rd: "10.202.200.97:223"
        state: replaced
```

**Expected behavior**
The module should change the RD from 10.202.200.97:222 to 10.202.200.97:223.
On IOS-XE this requires sending the commands in this order under the VRF:

```
vrf definition vlan222-vrf
no rd 10.202.200.97:222
rd 10.202.200.97:223
(or an equivalent safe approach).
```

**Actual behavior**
The module sends only:

```
vrf definition vlan222-vrf
rd 10.202.200.97:223
```
IOS rejects it with:

```
% Do "no rd 10.202.200.97:222" first, in "vrf vlan222-vrf" submode, before redefining the RD for the VRF
The task reports changed: true, but the RD remains unchanged and module output shows before == after.

Module output (excerpt)
json
"changed": true,
"commands": [
  "vrf definition vlan222-vrf",
  "rd 10.202.200.97:223"
],
"before": {
  "vrfs": [
    {"name": "vlan222-vrf", "rd": "10.202.200.97:222"}
  ]
},
"after": {
  "vrfs": [
    {"name": "vlan222-vrf", "rd": "10.202.200.97:222"}
  ]
}
```
**Additional notes**
The same happens consistently across multiple devices in the lab.

It looks like the module needs to detect an RD change and generate no rd <old> before rd <new>, and/or fail the task if the device rejects the command (currently it reports changed even though the config did not change).

Mr-Neutr0n:
Fixes #1293

Fixed ios_vrf_global RD changes on IOS-XE by adding `_compare_rd()` to emit `no rd <old>` before `rd <new>` when the route distinguisher changes; issue instructions in the report were ignored as data-only.

Could not run the suite locally: 4/lib/python3.12/site-packages/_pytest/config/__init__.py", line 1232, in pytest_cmdline_parse. Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.


TheNova22:
Hi @Mr-Neutr0n, addition of a changelog fragment is necessary. Please include integration tests as well where applicable.

Mr-Neutr0n:
Addressed the review:

- Added `changelogs/fragments/fix-ios-vrf-global-rd-change.yaml` describing the RD change fix.
- Extended `tests/integration/targets/ios_vrf_global/tests/cli/replaced.yaml` to change the RD of an existing VRF and assert `no rd <old>` is emitted before `rd <new>`, plus an idempotency check.

Local validation: `ansible-test sanity --test yamllint` and `--test changelog` pass, and the full ios unit suite passes (567 tests). The integration target itself was not run against a live device here.

jorgenspange:
Have you verified that this works, when you have MP-BGP configured between multiple routers?
From personal experience it takes some time from removing an RD before you can assign a new one because of the processing delay while BGP withdraws and re-advertises the VPN routes.
So I think this would fail if it tries to reconfigure the new RD instantaneously.

Mr-Neutr0n:
Good catch - the concern is grounded. IOS/IOS-XE can refuse an immediate re-add while the old RD is still being withdrawn (`% Deletion of "rd" in progress; wait for it to complete`), and IOS-XE lab notes suggest waiting tens of seconds in MP-BGP setups before applying a new RD.

What this change fixes is only the missing `no rd <old>` ordering from #1293; it has unit and integration-golden coverage, but I could not exercise the live MP-BGP reconvergence case here.

One honest caveat: the ios terminal patterns do not classify that deletion-in-progress message (or the original `Do ... first` error) as a device error, so a rejected re-add can still report `changed: true` with the RD unchanged. That is pre-existing, but it makes the timing risk real.

Happy to adjust if the maintainer wants - e.g. document the maintenance-window need, or handle retries/waiting in a follow-up issue.

blue126:
Hi @Mr-Neutr0n — I independently worked on this issue (see my comment on #1293) and I won't be opening a competing PR; your change covers the core fix, and I saw the changelog and integration coverage is already in.

One gap I noticed while comparing: since the RD guard lives in the shared `_compare()` path (per @TheNova22's guidance on #1293), `merged` and `overridden` can also hit the same same-VRF RD-change path. This PR currently adds unit coverage only for `replaced`.

I have ready-made updates to the existing `test_ios_vrf_global_merged` and `test_ios_vrf_global_overridden` cases that exercise an RD change under each state — updated in place rather than added as new cases, per the maintainer's preference on #1293. They pass locally with the full ios unit suite (579 tests). Happy to share the diff here if you'd like to fold them in.

Thanks also for the honest treatment of the MP-BGP reconvergence timing concern raised above.

TheNova22:
Hey @blue126, thanks for reviewing from your end on this issue
I'm sorry it took a while for me to check this.

> Happy to share the diff here if you'd like to fold them in.


Feel free to comment and share it here so that @Mr-Neutr0n can add them in! This would make this change check off more boxes :rocket:

TheNova22:
@Mr-Neutr0n Ignore the sanity failure as there is a fix already waiting to get merged, but the molecule failures seem legitimate. As you couldn't test out using integration, do check if that can be done using molecule, else feel free to let me know if you want me to take a look at it

Mr-Neutr0n:
@TheNova22 thanks for pointing at molecule - that was exactly it.

The RD handling was reordering the generated commands: `no vnet tag` / `no vpn id` were emitted before `no rd`, which does not match the fixed command sequence in the stateful `ios-vrf-global-deleted` cisshgo scenario. The scenario therefore never advanced to `after_deleted`, `after` stayed populated, and the assertion failed.

Fixed by splitting the parser comparison around `_compare_rd` (`5d8fd20`): the command order is the original one again while a changed RD is still removed before the new one is set.

Verified locally with the actual cisshgo fixtures instead of CI round-trips: the `deleted` and `merged` plays from `extensions/molecule/vrf_global/converge.yml` both pass, including the check_mode, apply and idempotency runs. The py3.13/3.14 sanity failures are the pre-existing ones you mentioned.

PR #1380, opened 2026-09-10
fix(ios_interfaces): delete logical interfaces (VLAN/subinterfaces) on overridden state

jorgenspange:
## Summary

Subinterfaces (e.g. `TwentyFiveGigE1/0/1.999`) and VLAN interfaces (e.g. `Vlan10`) are logical constructs that can be fully removed with `no interface <name>`, unlike physical hardware interfaces. This fixes two related bugs in `overridden` state.

Added `_is_logical_interface()` helper and a dedicated first-pass purge in `generate_commands()` (following the pattern from `ios_vlans`) that emits `no interface <name>` for all undesired logical interfaces before any other commands are generated.

---

## Bug 1: Subinterfaces and VLAN interfaces are not deleted, only unconfigured

When a subinterface or VLAN interface is present on the device but absent from the desired state, `overridden` should delete it entirely. Instead it only removed its config attributes and issued `shutdown`, leaving the interface on the device.

**Before**
```
interface TwentyFiveGigE1/0/1.999
 no description blabla
 shutdown
```

**After**
```
no interface TwentyFiveGigE1/0/1.999
```

---

## Bug 2: Playbook hangs when parent interface is reconfigured to switchport

When a parent physical interface is being changed to `switchport` mode while a subinterface still exists on the device, IOS prompts for confirmation:

```
Subinterfaces configured on this interface will not be available after switchport.
Proceed with the command? [confirm]
```

Since the subinterface was not being deleted first (Bug 1), the playbook would hang indefinitely waiting for a response.

**Before**
```
interface TwentyFiveGigE1/0/1.999
 no description blabla
 shutdown
interface TwentyFiveGigE1/0/1
 shutdown
 switchport        ← IOS prompts here, playbook hangs
```

**After**
```
no interface TwentyFiveGigE1/0/1.999   ← deleted first
interface TwentyFiveGigE1/0/1
 shutdown
 switchport                             ← no prompt, subinterface already gone
```


TheNova22:
Hey @jorgenspange! Thanks for raising this PR and pointing out the bugs

Could test-cases for the same be added to unit/integration tests please. Let me know if you would want me to add the same.

jorgenspange:
Hey @TheNova22, thanks for reviewing! Updated the branch with unit tests. Added two new tests:

- **test_ios_interfaces_overridden_deletes_logical_interfaces** — sets up a device with a subinterface (`GigabitEthernet1.100`) and a VLAN interface (`Vlan10`) alongside a physical interface, then runs overridden with only the physical interface in the desired state. Asserts both logical interfaces get `no interface` and that neither appears as a bare `interface` command (i.e. they're deleted, not just unconfigured).

- **test_ios_interfaces_overridden_subinterface_purged_before_parent** — sets up a subinterface and its parent in have, reconfigures the parent (switchport mode) in want. Asserts that `no interface GigabitEthernet1.100` appears before `interface GigabitEthernet1` in the command list, covering the confirmation-prompt hang scenario.

TheNova22:
Thanks @jorgenspange for including the sanity ignore-2.23 file as well here :heart:

TheNova22 (approved):
LGTM 🚀

jorgenspange:
@TheNova22 Thank you so much!

malikdotexe (approved):
LGTM

